Issue #7, opened 2026-05-02
[Feature]: remove legacy "bounty-agent" mentions


vmihalis:
### Problem

no bounty agent anymore

### Proposed solution

bob

### Confirmation

- [x] I did not include private target data, credentials, or unpublished third-party findings.

aGFja3RoZXBsYW5ldA:
rip

PR #6, opened 2026-05-01
Release v1.2.0: SC pipeline, capability-pack routing, adapter wrappers

ejmockler:
## Summary

`v1.2.0` bundles four cohesive changes:

- **Smart-contract testing pipeline** — six chain families (EVM, SVM, Aptos Move, Sui Move, Substrate ink!, CosmWasm), 21 SC MCP tools, hunter roles `hunter-substrate` and `hunter-cosmwasm`, an `evidence-agent` that dispatches by `surface_type`, SC-aware phase gates, and SS58/bech32 address normalization.
- **Capability-pack routing for hunter dispatch** — `mcp/lib/capability-packs.js` is the registry of routing decisions. `bounty_route_surfaces` classifies surfaces into packs and writes `surface-routes.json`; `bounty_start_wave` carries the route triple onto each assignment; `bounty_record_finding` persists it onto every finding; verifier/chain/evidence/reporter prompts dispatch on `finding.capability_pack` via a registry-rendered table. New `bounty_read_surface_routes` exposes the routes to downstream consumers. MCP tool bundles are per-chain; SC hunter agents went from 38 tools to 10–13. `role-model.js`, `claude-role-renderer.js`, `codex/role-specs.js`, `codex-role-renderer.js`, and `tool-registry.js` all derive their per-chain hunter entries from the registry, so adding a chain pack auto-extends every consumer.
- **Adapter auto-detection** — `hacker-bob install` works without `--adapter`. A layered detector picks the adapter based on reinstall metadata → host env (`$CLAUDE_PROJECT_DIR`, `$CODEX_HOME`) → project artifacts (`.claude/`, `.codex/plugins/`, `.agents/plugins/`, `.mcp.json`) → CLI on `PATH` → claude fallback. The chosen adapter and reason are logged to stderr.
- **Adapter wrapper packages** — new `hacker-bob-cc` and `hacker-bob-codex` thin wrappers each pin `--adapter <id>` as the default and delegate to the canonical `hacker-bob` CLI. `npx hacker-bob-cc install <dir>` and `npx hacker-bob-codex install <dir>` give adapter-discoverable install entry points. The canonical `hacker-bob` package and its `hacker-bob` binary are unchanged.

The README Quickstart also gained an `### For AI Coding Agents` subsection with an adapter-aware verification table plus a `## MCP Troubleshooting` section.

## Behavior changes (please review carefully)

When `--adapter` is omitted, four lifecycle commands now act differently:

[...]

Issue #4, opened 2026-04-26
Capture real sanitized first-run media


vmihalis:
Add real sanitized screenshots or short demos for the first-run docs. Capture doctor-ok.png, status-fresh.png, and hunt-start.png using docs/media/README.md as the checklist. Use only a throwaway Claude Code workspace and a controlled lab target that the maintainer owns or is explicitly authorized to test. Do not include credentials, private paths, real targets, reports, cookies, tokens, or session artifacts.

vmihalis:
Solved in PR #37.

PR #20, opened 2026-05-09
Add verification v2 adjudication flow

vmihalis:
## Summary
- add verification v2 runtime support for multi-round verification, context reads, adjudication, replay context schema tooling, and manifest/archive tracking
- wire verification state into MCP/session surfaces, role prompts, generated Claude/Codex skills, pipeline analytics, and export/status/debug surfaces
- harden freshness guarantees so stale adjudication source rounds, corrupted current-round coverage, and time-only auth expiry flags cannot weaken the verification pipeline
- add regression coverage for verification round writing, adjudication hashing/freshness, prompt contracts, install surfaces, and v2 snapshot stability

## Verification
- npm run check:syntax
- node --test --test-name-pattern='verification v2' test/mcp-server.test.js
- npm run test:mcp
- npm run test:prompts
- git diff --check
- npm test

## Summary by CodeRabbit

* **New Features**
  * Introduces verification schema v2: attempt-scoped snapshots, adjudication plans, and stricter v2 result/evidence binding.
  * Evidence packs bind to final verification hashes; replay context is supported for v2 replay/evidence flows.
  * Status and session views now include verification metadata (schema, attempt ID, snapshot/hash, freshness/stale blockers).

* **Chores**
  * Registered new verification tools and updated verification orchestration and manifest handling.


ejmockler:
## Seven follow-up commits resolving the review feedback

Pushed on top of `437a9d5`. All eleven CodeRabbit comments addressed, three regression tests added (canonicalization, manifest race, adjudication privilege). Full `npm test` green on a clean checkout.

### Why

**`1de2683` — fail-loud test fixture (CR #10).**
`seedVerificationPipeline` swallowed `readVerificationContext` errors and silently rewrote sessions to v1, so every downstream test inherited masked behavior. Now it rethrows on parse failure and bootstraps a real v2 VERIFY attempt when the test seeded a post-VERIFY phase without one. Surfaced fourteen callsites; three of them had evidence-pack assertions that were technically wrong but never failed. The helper was hiding regressions, not smoothing fixtures.

**`e7e7c43` — canonicalize v2 payloads (CR #3).**
v2's whole point is content-addressed verification: same inputs ⇒ same `final_verification_hash`. But `normalizeStringEnumArray` deduped without sorting, `normalizeArtifactHashes` preserved insertion order, and `writeVerificationRound` assembled `results` in caller order. Two semantically identical rounds written in different orders produced different bytes, different `input_round_hashes`, different `final_verification_hash`. Fix sorts at all three points. Two regression tests: permute `confidence_reasons`/`artifact_hashes` ordering and assert hash stability; reverse multi-finding input and assert deterministic on-disk bytes.

**`c99c9c2` — manifest rollback + analytics width + schema inference (CR #4, #5, #6).**
`writeSessionStateDocument` advanced `state.verification_attempt_id` before `refreshVerificationManifest` ran. A silent `{ok:false}` return left state pointing at an attempt the manifest could not describe and still emitted `phase_transitioned`. Now `throw_on_error: true` and a state rollback on failure. Also restored 128-char width for the three v2 hash fields in pipeline-event reads (write allowed 128, read truncated to 120 — silently changed identifiers analytics consumers saw). Also infers `schema_version: 2` from any current adjudication or v2 round, not just `snapshot.exists`.

**`4f9c072` — adjudication is orchestrator-only (CR #11).**
`bounty_build_verification_adjudication` was in the universal verifier bundle, giving brutalist/balanced/final verifiers permission to overwrite the adjudication artifact mid-round. Adjudication is the arbiter step between rounds — only the orchestrator should call it. `role_bundles → ["orchestrator"]`, `global_preapproval → false` per orchestrator-only-mutator policy, generators re-run, `.claude/settings.json` updated, prompt-contract assertion narrowed, and the tool added to the verifier-bundle forbidden list as a regression guard.

**`85144d5` — verifier prompt and renderer cleanup (CR #1, #2, #7, #8, #9).**
Scope `bounty_read_verification_round(round='final')` with `target_domain`; add an explicit `bounty_transition_phase(GRADE)` on the no-reportables VERIFY branch so the session never stalls; resolve the "exactly one call" vs retry contradiction in `evidence.md` ("exactly one *successful* write; do not retry on stale-final, retry on transient errors only"); pass `{ target_domain }` explicitly in renderer spawn templates; use `archived != null` consistently in pipeline event counts.

**`3efea66` — orphaned attempt recovery.**
`archiveCurrentV2Attempt` fell back to the literal string `attempt-unknown` when state had lost `verification_attempt_id` but artifacts remained on disk. Two consecutive recoveries collided. Now infers attempt_id from the snapshot file (which always carries it), then from round documents, and only as a last resort uses a `Date.now()`+`crypto.randomBytes(4)` suffix. Found while resolving CR #10 — the test fixture was triggering exactly this path. Production code now safe under partial-recovery scenarios.

**`6187205` — package test workspace tolerance.**
`test/package.test.js` walked `.claude/` for expected canonical files but `package.json`'s `files` glob ships `settings.json` only. Any developer with a Claude Code session that wrote `settings.local.json` failed `npm test`. Added it to `LOCAL_INSTALL_METADATA_FILES`. Pre-existing on `main`; bundled here because it would have been the next "I tried to run the tests" question.

### Implications

The cumulative effect is that v2 is now actually deterministic. `final_verification_hash` cited in a triage report can be reproduced byte-for-byte by anyone replaying against the same snapshot — not "approximately the same as long as nobody sorts differently." Manifest and state are now coupled: either both advance or neither. Adjudication is a privileged checkpoint only the orchestrator can mint, restoring the least-privilege story v2's architecture deserves. And the test suite is finally an honest signal — silent-v1-downgrade fixture behavior was masking the evidence-shape regression class, and removing that class of mask is the single highest-value change for this PR's long-term maintainability.

### Caveman closing

- Hash drift. Sort. Hash same.
- State ahead, manifest break, state stuck. Now state go back. No stuck.
- Many verifier touch judge tool. Pin to chief. Less mess.
- Two ghost archive same path. Read snapshot, find name. No crash.
- Test lie quiet. Test shout loud. Bug found.
- MCP no name. Add name. MCP happy.
- Local file break test. Skip file. Test true.

v2 now truly deterministic. Tribe trust hash. Hunt safer.

PR #100, opened 2026-06-12
Fix critical audit findings: command injection, unpinned dep, guard bypass

ai-13rac1:
## Summary

Fixes three critical/immediate findings from a security audit of the framework:

- **Fix command injection in `commandExists()`** — `scripts/install.js:287` interpolates its argument into a shell string. Added input validation against `/^[a-z0-9_-]+$/i` before reaching the shell.
- **Pin `@brutalist/mcp` to 1.14.7** — All adapters ran `npx -y @brutalist/mcp@latest`, fetching whatever is newest on npm on every invocation. Pinned to `1.14.7` (matching CI) and centralized the constant so all adapters import from `scripts/merge-claude-config.js` instead of defining duplicates.
- **Block on `shlex.split` `ValueError` in session guards** — Both `session-read-guard.sh` and `session-write-guard.sh` silently allowed commands when `shlex.split` raised `ValueError` on malformed input (e.g. unterminated quotes). Now both guards exit 2 (block) on parse failure.

## Test plan

- [x] Install smoke tests pass (containerized `node --test test/install-smoke.test.js`)
- [x] CLI and prompt-contracts tests pass (`node --test test/cli.test.js test/prompt-contracts.test.js`)
- [x] Hook guard tests pass (`python3 test/test-write-guard.py && python3 test/test-read-guard.py`) — 69/69 and 51/51, including new ValueError bypass test cases
- [ ] Pre-existing `doctor` command failures in container environment are unrelated (confirmed identical on `main`)

## Summary by CodeRabbit

* **Bug Fixes**
  * Hooks now reject malformed shell commands, emitting an error and exiting with status 2 to block unsafe execution.
  * Command-probing now validates command names to prevent shell injection.

* **Chores**
  * MCP server wiring consolidated to a shared configuration and pinned to `@brutalist/mcp`@1.14.7.
  * Centralized command-exists helper added and reused across scripts.

* **Tests**
  * Added tests for malformed-command blocking, command-name validation, and pinned MCP config.

### DIFF
--- a/.claude/hooks/session-read-guard.sh
+++ b/.claude/hooks/session-read-guard.sh
@@ -222,7 +222,12 @@ def check_bash_command(command):
     try:
         tokens = shlex.split(command, posix=True)
     except ValueError:
-        return
+        print(
+            "BLOCKED: Command cannot be safely parsed. "
+            "Refusing to allow potentially unsafe shell operation.",
+            file=sys.stderr,
+        )
+        raise SystemExit(2)
     for index, token in enumerate(tokens):
         command_name = pathlib.PurePosixPath(token).name
         if command_name not in READ_COMMANDS:
--- a/.claude/hooks/session-write-guard.sh
+++ b/.claude/hooks/session-write-guard.sh
@@ -205,7 +205,10 @@ def check_mutating_path_commands(command):
     try:
         tokens = shlex.split(command, posix=True)
     except ValueError:
-        return
+        block(
+            "BLOCKED: Command cannot be safely parsed. "
+            "Refusing to allow potentially unsafe shell operation."
+        )
 
     mutators = {"rm", "unlink", "mv", "cp", "chmod", "chown"}
     for index, token in enumerate(tokens):
--- a/adapters/codex/hacker-bob/.mcp.json
+++ b/adapters/codex/hacker-bob/.mcp.json
@@ -10,7 +10,7 @@
       "command": "npx",
       "args": [
         "-y",
-        "@brutalist/mcp@latest"
+        "@brutalist/mcp@1.14.7"
       ]
     }
   }
--- a/adapters/codex/index.js
+++ b/adapters/codex/index.js
@@ -8,6 +8,7 @@ const {
   updateCodexSkillFiles,
 } = require("../../scripts/lib/codex-role-renderer.js");
 const { createSafeInstallFs } = require("../../scripts/lib/install-fs.js");
+const { BRUTALIST_MCP_SERVER } = require("../../scripts/merge-claude-config.js");
 
 const id = "codex";
 const PLUGIN_NAME = "hacker-bob";
@@ -163,15 +164,6 @@ function managedDirs() {
     ".agents",
   ];
 }
-
-// External adversarial-roast MCP server consumed by the brutalist-verifier
-// role. Optional — registered alongside hacker-bob but not required at
-// runtime. See prompts/roles/brutalist-verifier.md for the graceful-fallback
-// contract.
-const BRUTALIST_MCP_SERVER = Object.freeze({
-  command: "npx",
-  args: ["-y", "@brutalist/mcp@latest"],
-});
 
 function mergeConfig({ serverPath }) {
   return {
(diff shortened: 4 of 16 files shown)


ejmockler:
## Maintainer review — adversarial pass (the one the skipped jobs would have done)

Heads up on why this PR shows two **skipped** checks: it's an external fork, so the `brutalist` and `Bob Diff Review` jobs are gated on `head.repo.full_name == github.repository` (forks don't receive org secrets — see the comment in `.github/workflows/bob-diff-review.yml`). So those AI reviews were skipped *by design*, not failing. I ran that adversarial review manually (5 independent lenses + a fork-malice scan). Summary below.

**Net: solid, security-positive PR — no malice, three genuinely correct fixes. One real regression, which I've fixed at the root in a follow-up commit on this branch (`75bd43e`).**

### Per-fix verdict
| Fix | Verdict | Notes |
|---|---|---|
| **Pin `@brutalist/mcp@1.14.7` + centralize the constant** | ✅ correct | `1.14.7` is the current `latest` (not a downgrade/typosquat/deprecation); single source of truth; require-safe (`if (require.main === module)` guard); the static codex `.mcp.json` is drift-guarded by a test that was verified to fire. |
| **Block on `shlex.split` ValueError in both guards** | ✅ correct | Both fail-closed (exit 2). In the write-guard, `block()` raises `SystemExit(2)` *before* the `for … in enumerate(tokens)` loop, so there's no unbound-`tokens` fall-through. Guard tests pass on this PR and fail on `main` — i.e. genuinely meaningful, not vacuous. |
| **Command-injection allowlist in `commandExists()`** | ✅ correct intent, ⚠️ had a regression (now fixed) | The regex neutralizes every injection vector (proven against a `touch`-sentinel payload — the shell never runs). Two issues, both addressed in `75bd43e` below. |

### Findings on the command-injection fix
- **F1 (regression, fixed):** `/^[a-z0-9_-]+$/i` excluded `.`, so `commandExists("jwt_tool.py")` (`install.js:300`, `lifecycle.js:305`) became permanently dead — a false `MISSING: jwt_tool` / doctor warning (advisory only, never a false-*allow*; the `~/jwt_tool/jwt_tool.py` fallback still covered the canonical layout). A `.` is shell-safe in `command -v <word>`, so the correct allowlist is `/^[a-z0-9._-]+$/i`.
- **F2 (defense-in-depth gap, fixed):** the guard was added to only *one* of the two duplicated `command -v ${command}` sinks. The twin in `adapters/index.js` (`defaultCommandExists`) was left unguarded — so "fix the injection class" wasn't actually repo-wide. (Not live-exploitable — all callers pass static literals — but the duplication is the root cause: it's *why* one copy got guarded and the other didn't.)

### What `75bd43e` does (root fix for F1 + F2)
- New `scripts/lib/command-exists.js` — the **single** allowlist-guarded `command -v` sink, regex `/^[a-z0-9._-]+$/i`.
- `scripts/install.js` and `adapters/index.js` both import it; the two local copies (and the now-unused `spawnSync` import in `adapters/index.js`) are removed. `install.js` re-exports `commandExists`, so `lifecycle.js` inherits the fix unchanged.
- `test/command-exists.test.js` pins both properties: injection payloads never reach the shell (sentinel-file proof), and a real dotted executable on `PATH` **is** detected (the `jwt_tool.py` regression guard — this test *fails* on the old `/^[a-z0-9_-]+$/i` and passes on the new one).
- Full suite green.

### Remaining (optional, non-blocking)
- **F3 (nit):** `cli.test.js:70` and `prompt-contracts.test.js:1707` now assert `=== BRUTALIST_MCP_SERVER.args`, i.e. self-referential against the constant the merge code derives from. Mitigated — literal `@1.14.7` anchors survive at `cli.test.js:98`/`:355`, so version drift is still caught. Keeping one literal anchor on the Claude-merge path too would close it. Not blocking.

### Malice check
Cleared. The diff is exactly the claimed files, every changed test assertion is equal-or-stronger, both new guard tests fail-on-main/pass-on-PR, the version pin moves toward the *current* latest (not an attacker-controlled/rollback version), and the guard change is fail-closed. The one regression is the signature of honest over-tightening (errs toward false-*negative*, never false-allow) — not a backdoor.

Thanks for the careful work on this — the three fixes are real and the guard tests are well-constructed. With `75bd43e` folded in, I'd consider this ready pending a fresh green CI run.

PR #109, opened 2026-06-14
feat(verify): web-path severity-rise guard + smart-contract carve-out (PR2)

vmihalis:
## What

Second PR in the offensive-Bob series (after #108's exploit-proof contract). Adds a **web-path severity-rise guard** to `writeVerificationRound`.

**The hole it closes:** a verification round sets each finding's `result.severity`, and today that value is only enum-validated — never compared to anything. On a **web** engagement, a verifier (over-eager or compromised) could raise a finding `low`→`critical` with **zero proof**, and the inflated severity flowed straight to the report.

**The guard:** on web sessions only, `result.severity` may exceed the frozen claim's severity **only** when backed by two-factor exploit proof:
1. `result.confidence_reasons` includes the new `exploit_replay_confirmed` reason, **and**
2. a frozen claim for that finding has an `exploit_run` evidence_ref satisfied by a real **MAC-signed** row in `offensive-runs.jsonl` (reuses the #108 `offensiveRunRowSatisfiesEvidence` / `readOffensiveRunRecords` primitives).

Otherwise the severity is **clamped** down to the frozen baseline (clamp, never reject — preserves round finding-coverage contracts). It runs inside `withSessionLock`, **before** the document build and the v2 `final_verification_hash`, so nothing inflated lands on the audit-graded path and the hash covers the clamped values.

## Scope decision (load-bearing)

The guard fires **only on web sessions** (`scope_policy.target_url != null`). Repo/OSS severity is independently capped at grade time; the **smart-contract** balanced-verifier is *supposed* to re-judge severity upward from on-chain effect — so the guard must not touch SC.

**Pre-merge review caught a real gap here and it's fixed in this PR:** `web`/`smart_contract` is a **per-finding** axis, and a web-scoped session can carry smart-contract surfaces in **cross-stack** mode (`web + smart_contract_evm`). The original session-scope-only check would have clamped those SC findings — the exact regression the design forbids. The guard now **carves out smart-contract findings** (detected via `surface_type` / `sc_evidence` / `capability_pack` stamped on the claim payload by `buildClaimPayloadFromFinding`). A web finding can never carry those signals, so pure-web guard strength is unchanged.

## Day-one behavior

No offensive runner writes rows yet, so `readOffensiveRunRecords` returns `[]` and **every unproven web rise clamps** — pure hole-closure. The exploit-backed *allow* branch is dormant by design and lights up automatically when a future runner PR (PR3) writes MAC-signed rows.

## Tests

`test/severity-rise-guard.test.js` — 11 cases on the **real** freeze/nucleus/round path (real `initSession`/`buildClaimFreeze`/`writeVerificationRound`, real `signOffensiveRunRow`): unproven-rise clamp, exploit-backed allow, both two-factor halves, **cross-stack web+SC carve-out (web clamps while SC stays raised)**, SC-via-`sc_evidence`-alone, equality/lowering/info/null/no-freeze, lazy-ledger tolerance (no-rise path never reads the ledger), v2 final+adjudication consume already-clamped rounds, and the three tool-schema enum copies.

- `npm run test:mcp` → **2421 pass / 0 fail** (2 skip)
- `npm run test:prompts` → 181 pass / 0
- `npm run test:hooks` clean · `check:syntax` clean · generators idempotent (no drift)

## Notes / honest boundaries

- **Cross-finding replay is intentionally out of scope.** A MAC-signed offensive row binds to its `(domain, target, command/hash)` fields, not to a `finding_id`. After a future runner PR ships, an agent with one genuine exploit row could in principle attach its public fields to another finding's `exploit_run` ref. That's a property of the **#108 contract + the future runner** (no agent-facing tool produces `exploit_run` refs today), not this guard — to be addressed where rows are first produced (bind row→finding).
- **Forward-compat note:** the exploit-backed allow path reads `result.confidence_reasons`, which only exists on v2 results; fresh web sessions are v2, and the path is dormant day-one regardless. A future runner PR should ensure v2 (or extend the signal) before relying on it.
- **Pre-existing flake (not introduced here):** the suite-wide `withTempHome` pattern mutates `process.env.HOME`; under the manifest's single-process concurrent run a rare cross-file race can transiently fail a `withTempHome` test. Observed once, not reproducible in repeated runs (4× clean full `test:mcp`); a re-run clears it. Worth a separate suite-wide hardening (`--test-concurrency=1` in the runner), out of scope for this PR.

## Release Notes

* **New Features**
  * Added `exploit_replay_confirmed` as an allowed verification confidence reason.
  * Enhanced “severity rise” guarding for web findings, including `severity_clamps` in verification outputs and a new `severity_clamped` pipeline event when clamping occurs.

* **Documentation**
  * Updated verifier v2 guidance and write-contract rules to require real exploit proof for web severity escalation and to govern when `exploit_replay_confirmed` is included.

* **Tests**
  * Added coverage for web vs non-web behavior, smart-contract special-casing, and v1/v2 verification flow integration.

### DIFF
--- a/.claude/agents/balanced-verifier.md
+++ b/.claude/agents/balanced-verifier.md
@@ -89,7 +89,7 @@ Each v1 `results` entry must include:
 - `reportable`: boolean
 - `reasoning`: required non-empty string
 
-For v2, add top-level `verification_attempt_id`, `verification_snapshot_hash`, and `round_profile: "balanced"` to the write call. Each result must also include `confidence`, `confidence_reasons`, `state_sensitive`, and `artifact_hashes`. Use the same allowed confidence reasons as brutalist; preserve `state_sensitive: true` whenever fresh state, auth, or chain state could change the outcome.
+For v2, add top-level `verification_attempt_id`, `verification_snapshot_hash`, and `round_profile: "balanced"` to the write call. Each result must also include `confidence`, `confidence_reasons`, `state_sensitive`, and `artifact_hashes`. Use the same allowed confidence reasons as brutalist; preserve `state_sensitive: true` whenever fresh state, auth, or chain state could change the outcome. Web severity rises above the frozen claim require real exploit proof; include `exploit_replay_confirmed` only when the replay is backed by the exploit-run proof contract.
 
 Do not write verifier markdown directly. The MCP tool owns `balanced.json` and the human/debug mirror.
 
--- a/.claude/agents/brutalist-verifier.md
+++ b/.claude/agents/brutalist-verifier.md
@@ -107,7 +107,7 @@ Each v1 `results` entry must include:
 
 For v2, the round must cover exactly the snapshot finding IDs and every `results` entry must also include:
 - `confidence`: `high|medium|low`
-- `confidence_reasons`: any of `fresh_replay_passed`, `auth_expired`, `tooling_blocked`, `state_changed`, `manual_inference`, `roast_disagreement`, `disambiguation_failed`, `agreement_not_replayed`
+- `confidence_reasons`: any of `fresh_replay_passed`, `auth_expired`, `tooling_blocked`, `state_changed`, `manual_inference`, `roast_disagreement`, `disambiguation_failed`, `agreement_not_replayed`, `exploit_replay_confirmed`
 - `state_sensitive`: boolean; set true when target state, auth state, chain state, or fresh replay timing could change the result
 - `artifact_hashes`: object of bounded replay/audit artifact hashes when available, otherwise `{}`
 
@@ -117,6 +117,7 @@ Suggested v2 confidence mapping:
 - Tooling/RPC blocked: include `tooling_blocked`, usually deny/fail closed unless local policy says otherwise.
 - Roast disagreement: include `roast_disagreement`.
 - Manual inference without replay: include `manual_inference`.
+- Web severity rises above the frozen claim require real exploit proof; include `exploit_replay_confirmed` only when the replay is backed by the exploit-run proof contract.
 
 Do not write verifier markdown directly. The MCP tool owns `brutalist.json` and the human/debug mirror.
 
--- a/.claude/agents/final-verifier.md
+++ b/.claude/agents/final-verifier.md
@@ -53,7 +53,7 @@ Each v1 `results` entry must include:
 - `reportable`: boolean
 - `reasoning`: required non-empty string
 
-For v2, add top-level `verification_attempt_id`, `verification_snapshot_hash`, `round_profile: "final"`, and `adjudication_plan_hash` to the write call. Every result must also include `confidence`, `confidence_reasons`, `state_sensitive`, and `artifact_hashes`; optional `inherited_confidence_reasons` and `resolved_confidence_reasons` are allowed.
+For v2, add top-level `verification_attempt_id`, `verification_snapshot_hash`, `round_profile: "final"`, and `adjudication_plan_hash` to the write call. Every result must also include `confidence`, `confidence_reasons`, `state_sensitive`, and `artifact_hashes`; optional `inherited_confidence_reasons` and `resolved_confidence_reasons` are allowed. Web severity rises above the frozen claim require real exploit proof; include `exploit_replay_confirmed` only when the replay is backed by the exploit-run proof contract.
 
 Do not write verifier markdown directly. The MCP tool owns `verified-final.json` and the human/debug mirror.
 
--- a/adapters/codex/skills/bob-evaluate/SKILL.md
+++ b/adapters/codex/skills/bob-evaluate/SKILL.md
@@ -1727,7 +1727,7 @@ Each v1 `results` entry must include:
 
 For v2, the round must cover exactly the snapshot finding IDs and every `results` entry must also include:
 - `confidence`: `high|medium|low`
-- `confidence_reasons`: any of `fresh_replay_passed`, `auth_expired`, `tooling_blocked`, `state_changed`, `manual_inference`, `roast_disagreement`, `disambiguation_failed`, `agreement_not_replayed`
+- `confidence_reasons`: any of `fresh_replay_passed`, `auth_expired`, `tooling_blocked`, `state_changed`, `manual_inference`, `roast_disagreement`, `disambiguation_failed`, `agreement_not_replayed`, `exploit_replay_confirmed`
 - `state_sensitive`: boolean; set true when target state, auth state, chain state, or fresh replay timing could change the result
 - `artifact_hashes`: object of bounded replay/audit artifact hashes when available, otherwise `{}`
 
@@ -1737,6 +1737,7 @@ Suggested v2 confidence mapping:
 - Tooling/RPC blocked: include `tooling_blocked`, usually deny/fail closed unless local policy says otherwise.
 - Roast disagreement: include `roast_disagreement`.
 - Manual inference without replay: include `manual_inference`.
+- Web severity rises above the frozen claim require real exploit proof; include `exploit_replay_confirmed` only when the replay is backed by the exploit-run proof contract.
 
 Do not write verifier markdown directly. The MCP tool owns `brutalist.json` and the human/debug mirror.
 
@@ -1861,7 +1862,7 @@ Each v1 `results` entry must include:
 - `reportable`: boolean
 - `reasoning`: required non-empty string
 
-For v2, add top-level `verification_attempt_id`, `verification_snapshot_hash`, and `round_profile: "balanced"` to the write call. Each result must also include `confidence`, `confidence_reasons`, `state_sensitive`, and `artifact_hashes`. Use the same allowed confidence reasons as brutalist; preserve `state_sensitive: true` whenever fresh state, auth, or chain state could change the outcome.
+For v2, add top-level `verification_attempt_id`, `verification_snapshot_hash`, and `round_profile: "balanced"` to the write call. Each result must also include `confidence`, `confidence_reasons`, `state_sensitive`, and `artifact_hashes`. Use the same allowed confidence reasons as brutalist; preserve `state_sensitive: true` whenever fresh state, auth, or chain state could change the outcome. Web severity rises above the frozen claim require real exploit proof; include `exploit_replay_confirmed` only when the replay is backed by the exploit-run proof contract.
 
 Do not write verifier markdown directly. The MCP tool owns `balanced.json` and the human/debug mirror.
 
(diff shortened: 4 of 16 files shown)


vmihalis:
@codex review

vmihalis:
@coderabbitai review

vmihalis:
## Review round 1 addressed — pushed `4a7d96b`

Thanks both. Summary of how each finding was handled:

**✅ SC carve-out was spoofable (brutalist Codex, high) — fixed.** You're right: deriving the exemption from the agent-settable claim payload let any web claim inject `sc_evidence` to dodge the clamp (my own test proved it). SC-ness is now resolved from the claim's `surface_ids` against the **write-guarded** `surface-routes.json` (`classifySurfaceCapability` output) via `readSurfaceRoutesStrict` — an agent cannot tamper that file. Pure-web sessions have no SC routes, so the carve-out never weakens them. New test `payload-injected sc_evidence cannot spoof the carve-out without a trusted SC route` locks it; the cross-stack test now seeds real trusted routes. This also resolves the inverse finding (legacy/minimal-payload SC claims falsely clamped) — detection no longer depends on payload presence.

**✅ Nucleus failed open (CodeRabbit major + Codex P2) — fixed, with nuance.** A `session-nucleus.json` that **exists but fails to parse** now fails **closed** (`STATE_CONFLICT`) — the guard can't be silently disabled by tampering the nucleus. An **absent** nucleus whose state is also unreadable stays pass-through (indeterminate scope / partially-seeded session), because a blanket fail-closed broke legitimate minimal sessions (`seedFinalVerification`-style). A real web session at VERIFY always derives a `target_url` nucleus from its readable state, so it never lands in the indeterminate branch.

**✅ Silent clamp / no audit trail (brutalist Claude high + Codex medium) — fixed.** The tool response now returns a `severity_clamps: [{finding_id, from, to}]` array, and a new bounded `severity_clamped` pipeline event persists the same per-finding detail (new event type + `clamps` field in `pipeline-events.js`). A runtime clamp is now distinguishable from a verifier's choice in a persisted artifact.

**✅ `VERIFY_SEVERITY_RANK` hand-typed island (brutalist Claude medium) — addressed.** Added a test asserting every `SEVERITY_VALUES` entry (and `informational`) maps to a non-zero rank, so a new tier can't silently disable the clamp.

**↩️ Corrupt freeze blocks verification (brutalist Claude high) — already the case upstream; not a new regression.** A corrupt `claim-freeze.json` is rejected *before* this guard runs — v1 via `findingIdSetForVerificationContext`, v2 via the snapshot freshness check `assertSnapshotMatchesFreeze` (both read the freeze earlier in `writeVerificationRound`). I confirmed this by trying to write a freeze-fail-closed test: the throw originates upstream, not in the guard. The guard's own freeze read is wrapped fail-closed purely as defense-in-depth (documented in-code).

**↩️ Freeze trusted without re-validating `freeze_hash` (Codex P2) — inherits the existing trust model.** `claim-freeze.json` is the authoritative frozen baseline the **entire** verification pipeline trusts (finding-id set, snapshot, adjudication, grade), and for v2 the snapshot freshness check already re-reads it. Adding a freeze-hash recompute only in this guard wouldn't change the model. Tampering the freeze requires write access to an MCP-owned session artifact — the same in-process-code-execution boundary documented for #108's HMAC ledger (the offensive-sandbox PR carries the privilege-separation requirement). Happy to add a pipeline-wide freeze integrity check separately if you'd prefer.

Gates: `test:mcp` 2424/0, `test:prompts` 181/0, `test:hooks` 75/75 + 55/55, generators idempotent.

@codex review
@coderabbitai review

vmihalis:
r2 pushed `d1e74cc`: CodeRabbit's read-path finding fixed — `normalizePipelineEventForRead` now mirrors `normalizeClampList` so the `severity_clamped` audit trail survives the read path, with a write/read round-trip test. Codex r1 had no new findings. test:mcp 2425/0, test:prompts 181/0.

@codex review
@coderabbitai review

vmihalis:
## Round 2 addressed — pushed `424147f`

- **`normalizeClampList` read-path integrity (Claude, correctness):** fixed — `from`/`to` are now rejected unless they're in `SEVERITY_VALUES`, on both write and read paths, so a tampered events log can't surface a bogus `supersafe`/`DROPPED` transition to operators. Test covers it.
- **In-place mutation contract (Claude, maintainability):** fixed — renamed `clampUnprovenSeverityRises` → `applyUnprovenSeverityClamps` and documented that it mutates `results` in place (load-bearing: the same array is serialized into the persisted round doc).
- **`exploit_replay_confirmed` proves claim-time proof, not the asserted severity (Codex, security):** valid and important — documented in-code as a hard requirement of the future offensive-runner PR rather than fixed here. The allow-path is a *floor* (a real MAC-signed exploit row + verifier assertion), raising the bar from zero-proof; proving the *asserted severity* requires the runner to bind each row to the demonstrated impact and anchor it to the current verification window (e.g. a `verification_exploit_run` ref kind), which the guard would then check. The branch is **dormant today** (no runner writes rows, so `readOffensiveRunRecords` returns `[]` and every web rise clamps), so there is no live exposure — it activates only once the runner exists, at which point the binding lands with it.

test:mcp 2425/0, test:prompts 181/0.

@codex review
@coderabbitai review

vmihalis:
## Round 3 addressed — pushed `0655824`

Both critics converged on three concerns; all fixed:

- **Exploit proof not bound to the asserted severity (Codex + Claude, primary gap):** the allow-path now requires the matching MAC-signed row to carry a `demonstrated_severity` that meets or exceeds the asserted severity. A low-impact row (e.g. a reflected canary) can no longer unlock a `critical` rise. No producer writes `demonstrated_severity` yet, so `verifySeverityRank(undefined) === 0` and the branch is **inert today** (every web rise clamps) — it activates only when the offensive-runner PR ships impact-bound rows. New test: `an exploit row that demonstrates a lower severity does not unlock a higher rise`.
- **Semantic nucleus drift (Codex):** scope is now derived from the **validated session state** (`sessionNucleusFromState(readSessionStateStrict(...).state)`), not the non-write-guarded `session-nucleus.json`. A corrupt or drifted nucleus file can no longer disable the guard on a session whose state still authorizes a web target. New test.
- **Unreadable routes silently clamped SC (Codex + Claude):** `smartContractSurfaceIdsFromRoutes` now fails **closed** (`STATE_CONFLICT`) when `surface-routes.json` exists but is malformed — clamping an SC finding would violate the invariant. Genuinely-absent routes stay pure-web (no SC surfaces). New test.

Remaining runner-PR requirement (documented in-code): the offensive runner must also anchor each row to the current verification window (replay freshness), in addition to setting `demonstrated_severity`.

test:mcp 2427/0, test:prompts 181/0, test:hooks 75/75+55/55, generators idempotent.

@codex review
@coderabbitai review

vmihalis:
## Round 4 addressed — pushed `673b8a8`

The recurring theme across rounds was that the SC carve-out kept depending on the next agent-controllable link (payload → surface_ids). Round 4 removes that dependency entirely:

- **SC carve-out spoofable via `claim.surface_ids` (Codex P1):** `surface_ids` is agent-supplied (the no-wave `bob_record_candidate_claim` path doesn't validate surface assignment), so a web claim could cite a real SC surface_id to dodge the clamp. **Removed all per-finding SC detection.** The clamp now scopes to **pure-web sessions**: if the session routes *any* smart-contract surface — per the trusted, **write-guarded** `surface-routes.json`, which an agent cannot add or remove — the clamp is skipped session-wide. This is spoof-proof (session-level, no per-finding trust), simple, and a **strict, non-regressing addition**: cross-stack sessions behave exactly as they did before this guard existed; pure-web (the common case) is fully guarded. New tests cover both directions, including the surface_id-citation P1 vector being clamped in a pure-web session.
- **Clamp audit not in the durable artifact (Codex + Claude):** the clamp is now recorded in the persisted, **content-hashed** round document as `severity_clamps: [{finding_id, from, to}]`, preserved + enum-validated by the read normalizer so the v2 `final_verification_hash` round-trips. The pipeline event is reduced to an authoritative `counts.clamped` (no per-finding list → the count/list divergence you flagged is gone). The tool response keeps `severity_clamps` too.
- **Intra-claim exploit-row spillover (Codex):** documented in-code as a runner-PR binding requirement (same as the cross-finding-replay note); the path is dormant today.
- **State-read pass-through observability (Codex + Claude):** documented — this branch is ~unreachable in production (the write tool authority validates state.json before the handler), and the durable clamp audit now lives in the content-hashed document regardless.

test:mcp 2428/0, test:prompts 181/0, test:hooks 75/75+55/55, generators idempotent.

@codex review
@coderabbitai review

vmihalis:
## Round 5 addressed — pushed `75108d2` (a design change, operator-approved)

The recurring SC-carve-out findings were a structural signal, and round 5's Codex P1 confirmed it: **the smart-contract carve-out cannot be made spoof-proof.** Every per-finding *and* per-session SC signal available before VERIFY is agent-influenced — `surface_ids`/payload are agent-recorded, and even `surface-routes.json` can be influenced (an evaluator can inject a synthetic `smart_contract` `surface.observed` via `bob_append_frontier_event`, which routing then classifies). A spoofable exemption is an inflation bypass — strictly worse than none.

**Resolution: dropped the SC carve-out. The clamp now applies uniformly to every finding in a web-scoped session** (the init-time `scope_policy.target_url` is the only trusted, non-agent signal). A smart-contract finding in a cross-stack session is held to its frozen (evaluator-recorded) severity unless exploit-proven — the evaluator's assessment is the trusted baseline, and an unproven verification-time raise above it is clamped exactly as for web. Pure smart-contract engagements are repo-scoped and untouched. This removes the entire per-finding/route SC attack surface and matches Codex P1's ask ("keep web clamps despite SC surfaces").

Also fixed:
- **`exploit_replay_confirmed` without a validated rise (Codex):** the proof check now runs independent of whether severity rose, and the reason is **stripped** from the persisted result when it didn't back a validated rise — so a clamped/non-rise result never carries a false proof signal. Other reasons are preserved.
- **Cross-finding exploit spillover (Claude):** `exploit_run` refs are bound only to single-finding claims; a multi-finding claim can't let one row prove a sibling's rise.
- **Broad state-read catch (Claude):** now distinguishes "no state file" (pass through) from "state exists but unreadable" (fail closed).
- **Mutation naming (Claude):** renamed to `clampResultSeveritiesInPlace`.

test:mcp 2428/0, test:prompts 181/0, test:hooks 75/75+55/55, generators idempotent.

@codex review
@coderabbitai review

vmihalis:
## Round 6 addressed — pushed `7416e4a`

These tightened my own round-5 strip:
- **`exploit_replay_confirmed` not stripped from `inherited_`/`resolved_confidence_reasons` (Codex HIGH):** correct — my round-5 strip only cleared `confidence_reasons`, but the schema permits the proof claim in all three arrays. The strip now **triggers when the reason appears in any of the three arrays and clears it from all of them** (`REASON_ARRAY_FIELDS`) unless it backed a validated rise. Test extended to cover the provenance arrays.
- **`demonstrated_severity` enum trap (Claude):** documented that the future offensive runner must write `demonstrated_severity` as a valid `SEVERITY_VALUES` member before signing — an unrecognized value ranks 0 (fail-safe no-unlock, but unobservable). Folded into the existing runner-requirements comment.
- **Strip not in schema docs (Claude):** added a description on the `confidence_reasons` field explaining `exploit_replay_confirmed` is preserved only when it backs a validated exploit-proven rise, else stripped.

test:mcp 2428/0, test:prompts 181/0, generators idempotent.

@codex review
@coderabbitai review

PR #110, opened 2026-06-14
feat(offensive): bob_http_confirm read-only confirmer (negative-only; signed producer deferred)

vmihalis:
## What

Adds **`bob_http_confirm`** — a trusted, read-only confirmer that is the **first signed `offensive-runs.jsonl` producer**, making the un-fakeable exploit-proof allow-path from #108 (HMAC proof contract) and #109 (web severity-rise clamp) reachable **end-to-end** for the first time.

It runs a read-only, same-endpoint **authorization differential** against a **server-minted synthetic id** and, only when the target genuinely exposes a resource that the real id auth-gates, writes a **MAC-signed `demonstrated_severity:"low"` row**. A verifier can then raise a finding to LOW *only* when backed by that real, signed proof.

## The contract (how it can't be gamed)

- **confirm-then-record, surface-anchored.** Confirm takes a `surface_id` (the finding doesn't exist yet), derives the request server-side from the surface's recorded route + a `{id}` template, mints the synthetic id itself. The row carries **no `finding_id`**.
- **`run_id` single-use.** `bob_record_candidate_claim` rejects a claim citing a `run_id` already consumed by a prior claim — read + checked under the same `withSessionLock` as the append. This is a **server-asserted 1:1 row→finding binding**, stronger than a row-stamped id, and works with the real ordering (the row is signed before the finding is minted).
- **`demonstrated_severity` is a true ceiling, not just a rise-limiter.** Hardcoded `low` by trusted code (a synthetic read proves a missing *gate*, not impact). The **record gate rejects** an `exploited_safely` claim asserting above it, and **verify clamps** any exploit-backed finding down to it even with no rise.

## Read-only is fail-closed CODE

- **`{id}` must be the final path segment** — structurally forbids an action/sub-resource segment after `{id}` (`/accounts/{id}/transfer`, `/payments/{id}/capture`), which is where mutation verbs live and where the baseline request would hit the **real** recorded id. A verb denylist can never enumerate the mutation surface, so the position rule is the boundary; a *narrow* destructive-verb denylist (with recursive percent-decode) is best-effort defense-in-depth for verb-named collections before `{id}`.
- Method allowlist (GET/HEAD/OPTIONS), header allowlist (strips method-override family), no request body, `follow_redirects:false`, no query string, scope re-asserted per hop.
- **Classifier** requires an *affirmative* resource shape: rejects soft-404 / error-envelope / empty / empty-paginated-collection JSON, SPA/app shells, and XML status-envelopes; **fails closed on unknown/missing content-type**; HEAD/OPTIONS can never mint a row. Never classifies on "not 401/403" alone.
- Row integrity: the per-row HMAC covers every consumer-trusted field; `validateNoSensitiveMaterial` runs before signing; the value-blind target drops the query string; the `offensive_outcome` label is asserted against the scored classification.

## Review

Reviewed adversarially with a multi-agent workflow (6 lenses) plus two re-verification passes; an initial blocker (read-only path hole) and the classifier false-positive class were found and fixed, and each fix was itself re-attacked in both directions (closure + no new false-negatives). Remaining residuals are all **LOW-capped, fail-closed, or deferred by design** and documented in code:

- verb-named collection *before* `{id}` (`/capture/{id}`) — the documented "shrinks, not eliminates" residual;
- sub-resource reads out of scope by the `{id}`-last rule (deferred to a future oracle that synthesizes its own baseline);
- generic-config-JSON precision residual; NDJSON / bare-scalar / binary-resource recall (fail-closed).

**Honest boundary:** PR3 does not give *absolute* un-fakeability — the per-session signing key is readable by an agent executing same-UID in-process code; that is the deferred offensive-sandbox PR (same assumption as wave-handoff signing). The MAC raises forgery from `printf >> ledger` to "import Bob and steal the key."

Two non-blocking latent-hardening follow-ups noted for later: the `claim_id` invariant coupling in the run_id binding, and source-side scope gating in `resolveSurfaceOrigins`.

## Tests

`npm test` green locally (test:mcp 2442 pass / 0 fail; test:prompts, test:hooks, check:syntax clean). New coverage: confirm→claim→freeze→verify `info→low` rise; unbacked claim rejected; `run_id` single-use; severity ceiling (record + verify); `{id}`-last structural boundary; read-only verb rejection; classifier precision/recall matrix (soft-404, empty-paginated, SPA, XML envelope vs. genuine large/XML resources, unknown content-type fail-closed).

## Summary by CodeRabbit

* **New Features**
  * Added a new HTTP endpoint confirmation tool for validating vulnerability status through differential analysis.
  * Enhanced exploit evidence validation with single-use enforcement and freshness checks.

* **Bug Fixes**
  * Strengthened severity claim validation to prevent claims from exceeding demonstrated capability.
  * Added matching requirements for exploit run evidence backing.


vmihalis:
@codex review

vmihalis:
@coderabbitai review

vmihalis:
Thanks — addressed all findings from the Codex + brutalist pass in `74bf565`:

**Security / correctness**
- **P1 symlink (write-side):** added write-side containment for the offensive-runs sink (mirrors the #108 read guard) — refuses a symlinked dir/ledger/run-file before writing, inside the session lock.
- **Severity regression:** removed the unconditional verify-time demonstrated-severity ceiling — it wrongly capped a finding whose higher baseline came from a *separate* non-exploit claim. The record-gate ceiling + rise-guard already cover the exploit-backed case; added a regression test.
- **Transport failures:** return `failure_reason` instead of `error`, so `executeTool` surfaces them as the intended `blocked_by_infra` negative confirmation rather than `INTERNAL_ERROR`.
- **Egress:** use the session's bound egress profile instead of hardcoding `default` (confirm was unusable on regional/proxy sessions).
- **Encoded separator:** reject a percent-encoded `/` after `{id}` (`/{id}%2Fcapture`, single + double encoded).
- **Circuit breaker:** append `http-audit.jsonl` records for both confirm requests so the session budget/breaker counts confirm traffic.

**Precision**
- Status/health objects (`{ok:true}`, `{success:false}`, `{service,region}`) and empty paginated collections (`{items:[],total:0}`) are no longer "resource-shaped".
- Strip a recorded endpoint's query so the baseline/target differential isolates id/auth.

**Hygiene**
- `mutating: true` (it appends the audit-graded ledger); redact sensitive response headers in `.poc.json`; fail closed on an unrecognized claim severity in the proof gate.

Residual (documented, low-capped): a catch-all that returns `{id:"default",...}` for any id still reads as a resource — robustly fixing that needs a second-synthetic-id differential, noted as a follow-up.

`npm test` green (test:mcp 2442 / test:prompts 181 / test:hooks 55).

@codex review
@coderabbitai review

vmihalis:
Pushed `1b2c58e`: regenerated the authority inventory (`docs/refactor-authority-inventory.md`) so `check:authority-inventory --check` matches the registry — the earlier `74bf565` flipped `bob_http_confirm` to `mutating:true` (it appends the audit-graded `offensive-runs` ledger) but I'd run the focused test commands locally rather than the full `npm test`, so the generated inventory was stale and CI's `test (20)/(22)` failed on the gate (not on a test). No logic change in this commit — one generated doc row (`mutating=false`→`true`). Full `npm test` is now green locally.

@codex review
@coderabbitai review

vmihalis:
## Direction change: `bob_http_confirm` is now NEGATIVE-ONLY (`16027fd`)

An adversarial review of the round-1 catch-all fix surfaced a **blocker that invalidated the whole positive path**, so I've reworked it rather than patch around it.

**The flaw:** the confirmer probes a *synthetic non-existent* id (`bob-synthetic-nonexistent-<uuid>`). A resource-shaped `200` on such an id can **only** mean a catch-all / server variance — never a sound per-object exposure. A deterministic per-object store `404`s an unknown id; a deterministic catch-all serves *both* synthetic ids. **No deterministic endpoint can answer two equivalent non-existent ids differently**, so the second-synthetic "discriminator" write-row branch was reachable *only* via server non-determinism (caching, rate-limit, LB heterogeneity) — which would then be **MAC-signed as exploit proof**. The un-fakeable contract signing noise is worse than no producer.

**The fix:** make the confirmer **negative-only** — it runs the read-only differential and reports a diagnostic outcome, but **never mints a signed `offensive-runs.jsonl` row**. The sound signed-row producer (a real **second-test-identity IDOR oracle** — `bob_auto_signup`, no synthetic-id guesswork) is a deferred follow-up PR.

**What's in this commit:**
- Removed the producer machinery from the confirmer: `buildSignedOffensiveRunRow`, capture artifacts, the write-side symlink sink-guard, the second-synthetic probe, the whole write path — plus their now-orphaned imports/exports (verified zero dangling refs).
- `classifyDifferential` is purely negative: a resource-shaped synthetic target → diagnostic `synthetic_id_resource_shape_not_provable`, never a `write_row`.
- Descriptor: negative-only description; `session_artifacts_written → ["http-audit.jsonl"]` (the only artifact it writes, via `appendHttpAuditRecord`); `mutating:true` retained (writes http-audit like `bob_http_scan`); dropped the unused `replay_context` input.
- Baseline-fetch failures now audited against the in-flight URL (was the round-1 nit).
- Regenerated authority inventory.

**The #108 proof contract is unchanged** (record gate, `run_id` single-use binding, severity ceiling) and stays proven by seed-based unit tests (hand-written signed rows) in `test/offensive-proof-contract.test.js` + `test/severity-rise-guard.test.js`, plus a seed-based `claim→freeze→verify` test in the confirmer suite.

A focused opus adversarial reviewer confirmed: no reachable write path, classifier purely negative, zero dead code, descriptor honest, fail-closed error path, contract proven 3 ways. Full suite green: `test:mcp` 2443/0, `test:prompts` 181, `test:hooks` 55/55, syntax + authority-inventory clean.

@codex review
@coderabbitai review

vmihalis:
## Review-thread disposition (`6accf8f`)

Triaged every open thread. Three live findings on the negative-only code are **fixed in `6accf8f`**; the rest are **moot** (the negative-only rework deleted the row-minting they warn about) or **deferred to the signed-producer PR** (dormant — nothing writes rows/captures here).

### Fixed in `6accf8f`
- **P1 — `scope_decision: null` dropped the http-audit append** (`bob-http-confirm.js:40` / `offensive-confirmer.js:577`): `normalizeHttpAuditRecord` requires a non-empty `scope_decision`; the null threw and was swallowed, so successful probes wrote **no** audit record. Now `"allowed"` on success / `"request_error"` on transport failure (matches `bob_http_scan`). Test added (asserts both probes land as `"allowed"`).
- **P1 — encoded separator in the baseline id** (`:208`): `pathTemplateMatchesEndpoint` now requires the percent-**decoded** pathname to also match a single `{id}` segment, so `/api/accounts/known%2Fdelete` is rejected. Test added.
- **medium — `assertAllowedRequestHeaders` dead self-check** (`:116`): removed (validated `HEADER_SUBSET` against itself; `assertNoForbiddenInputs` already blocks caller headers).

### Already fixed earlier on this branch
- query symmetry (`:299`, baseline strips recorded query + template rejects `?`), encoded-after-`{id}` in the template (`:178`, recursive decode), egress profile, transport `failure_reason`, verification-round capping regression.

### Moot under negative-only (no signed row is ever minted)
The classifier false-positive threads — catch-all template objects, `{ok:true}`/status JSON, JSON-wrapper recursion, text-denial pages, OPTIONS bodies (`:425/:423/:475/:509` and the brutalist `:425`) — all warned that `isResourceShapedResponse` could **mint a signed `exploited_safely` row**. It no longer can: the classifier is purely diagnostic and there is no write path. A misclassification only changes a negative `reason` string. Same for the symlink/hardlink sink-guard, mutating-proof-writer, and second-synthetic threads — that code was removed.

### Deferred to the signed-producer PR (valid but dormant here)
- **`claims.js:930` (P1) — bind exploit rows to the claim surface**: real defense-in-depth, but no producer writes rows in this PR, and rows require the HMAC key (the #108 bedrock), so an agent can't forge cross-surface reuse today. The strong binding (`run_id` single-use at the record gate) is already in place; surface-binding belongs with the producer that establishes row↔surface↔finding semantics. Tracked.
- **`paths.js:358` (P2) — add `offensive-runs/` captures to the write guard**: the negative-only confirmer writes no captures, so there is nothing to guard yet. Lands with the producer that writes them.

@codex review
@coderabbitai review

vmihalis:
## Fixes for the `6accf8f` review (`2e5d2d3`)

The negative-only confirmer still makes live probes, so the request-safety findings are valid. Fixed:

- **P1 — `{id}` action suffixes + deep-encoded separators** (`offensive-confirmer.js:177` / `:68`): replaced the suffix check with an allowlist — nothing may follow `{id}` except an inert file extension. One rule now rejects a following segment, an encoded separator at **any** depth (`%2525252F`), and same-segment action/matrix suffixes (`{id}:capture`, `{id};delete`). The baseline matcher rejects any-depth encoded separators via a layer-independent regex. Tests added.
- **P1 — replay safety** (`bob-http-confirm.js:26`): re-added `replay_context` to the schema and mapped `bob_http_confirm` to the web pack's replay policy, so a verifier's confirm probes are leased like `bob_http_scan`. (Removing `replay_context` earlier was my mistake — it's the replay-lease key, not just a row-stamp.)
- **medium — bare `Error`** (`record-candidate-claim.js:379`): `normalizeExploitRunEvidenceRefs` now throws `ToolError(INVALID_ARGUMENTS)` with machine-readable codes instead of surfacing as `INTERNAL_ERROR`.

**Deferred (documented in `claims.js`): the severity-ceiling row↔claim-surface binding** (`claims.js:975`). It's **dormant** under negative-only — no producer writes rows, so `readOffensiveRunRecords` is empty and no `exploited_safely` claim can even be backed; and the claim's surface doesn't yet flow to the gate (`record-candidate-claim` passes a singular `surface_id`). The finding/surface binding lands with the signed-producer PR that gives rows a real surface and wires it through. The existing `run_id` single-use + content-bind hold today.

@codex review
@coderabbitai review

vmihalis:
Round-5 fixes (`c0bb9b4`), made **comprehensive** to converge the class:

- **id-segment safety** (`:64`,`:202`, split-hex `%25%32%46`): one rule (`capturedIdSegmentIsSafe`) on the captured recorded id — rejects a path separator literal/encoded at **any** depth (decode-to-fixed-point) AND action/matrix punctuation (`: ; ,`) literal or escaped, fail-closed on a remnant escape. Replaces the regex-only check that missed split-hex + surface-record matrix punctuation.
- **inert-extension allowlist** (`:189`): closed set (`.json/.xml/...`) so `{id}.capture` is rejected while `{id}.json` passes.
- **bare Errors** (`:392`,`:489`): `normalizeExploitRunEvidenceRefs` + `normalizeExploitOutcome` now wrap the deeper claims.js validators as `ToolError(INVALID_ARGUMENTS)`.

8-variant rejection test + verb-dot-suffix + malformed-`exploit_outcome` tests added. `test:mcp` 2445/0.

@codex review
@coderabbitai review

vmihalis:
Round-6 disposition (`b24ca1d` fixes + `ccfc70b` docs) — **merging with documented residuals** (negative-only confirmer; produces nothing):

**Fixed (`b24ca1d`):**
- `:58` denylist now matches verb + format/matrix suffix (`/api/delete.json/{id}`)
- `:142` mutation check runs on the decoded path (`/api/%6Dutation/{id}`)
- `:599` swallowed http-audit failures now log to stderr
- `:243` intentional `blockInternalHosts:false` is now a named `SCOPE_VALIDATION_OPTS` constant (SSRF enforced at fetch)

**Documented residuals (`ccfc70b`) — no clean fix / safe / dormant:**
- `:214` dot-action suffix in a RECORDED id: unfixable without false-rejecting legit dotted ids (`user.name`, `1.2.3`); template side is closed. Documented in `capturedIdSegmentIsSafe`.
- `:331` query-routed endpoints: a safe **false-negative** (the confirmer never mis-confirms one, just skips it). Documented + noted in the tool description.
- `claims.js` cross-finding laundering: **dormant** (no producer writes rows) → tracked in #111, which gates the signed-row producer PR.
- `:396` P3 secret-shaped ref: a pre-existing `claims.js` sensitive-material bare-Error pattern, out of scope here.

CI green every round; `test:mcp` 2445/0. Merging.

Issue #111, opened 2026-06-15
Producer PR: bind exploit rows to the claim surface (severity-ceiling cross-finding gate)


vmihalis:
Tracking the deferred surface-binding gate flagged on PR #110 (Codex + brutalist, `claims.js` severity ceiling).

## Gap
`assertExploitedClaimHasProof` enforces a severity ceiling (`claimRank <= maxDemonstratedRank` over backed rows) plus `run_id` single-use, but it does NOT require a cited `exploit_run` row's `surface_id` to match the claim's finding/surface. So once a signed-row PRODUCER exists, a claim for surface A could cite a higher-severity row produced for surface B (same run_id, used once) and the ceiling would pass — cross-finding severity laundering.

## Why it's dormant today
PR #110 ships `bob_http_confirm` as **negative-only** — it writes no `offensive-runs.jsonl` rows, so `readOffensiveRunRecords` is empty and no `exploited_safely` claim can be backed at all. Rows also require the per-session HMAC key, so an agent can't forge cross-surface reuse. The claim's surface also doesn't yet flow to the gate (`record-candidate-claim` passes a singular `surface_id`, not `surface_ids[]`).

## Required before any signed-row producer ships
1. Wire the claim's surface through to `assertExploitedClaimHasProof`.
2. Require each backed `exploit_run` row's `surface_id` to be in the claim's surfaces (additive to `run_id` single-use + content-bind).
3. Add a test in `test/severity-rise-guard.test.js` that fails if a high-severity row for surface B can back a claim for surface A.

Referenced in code at `mcp/lib/claims.js` (severity-ceiling comment).

PR #16, opened 2026-05-05
add optional recon tool coverage

aGFja3RoZXBsYW5ldA:
## Summary

- Add optional Katana crawling to normal and deep recon so bounded in-scope crawl URLs augment CDX/Wayback and JS extraction.
- Add JWT-shaped candidate extraction from JavaScript artifacts plus jwt_tool availability checks for later authorized token review.
- Add deep-mode dnsx/tlsx enrichment and bounded subzy checks for CNAME takeover candidates.
- Update installer, doctor, docs, generated Claude/Codex artifacts, and prompt-contract coverage for the new optional recon tools.

## Validation

- npm run check:syntax
- node scripts/generate-claude-roles.js --check
- node scripts/generate-codex-skills.js --check
- npm run test:prompts
- node --test test/install-smoke.test.js
- node --test test/cli.test.js

PR #119, opened 2026-06-16
feat(offense): bob_http_idor_confirm second-test-identity IDOR signed producer (inert)

vmihalis:
## What

PR-C of the offensive sequence: **`bob_http_idor_confirm`**, the first *sound* signed-row producer. Identity B writes a producer-minted **256-bit crypto canary** into B's own object `O_B` at create time; the positive fires **only** when identity A (a different tenant) reads `O_B` and the response carries B's exact canary **at the same parsed leaf** — never a substring, never whole-body equality. A 256-bit nonce a deterministic/reflective server never saw is the categorical escape from #110's non-existent-id trap.

A signed `offensive-runs.jsonl` row is minted only when **all** mint conditions hold across a 7-probe **P0–P6** differential (P4/P4-id auth-challenged, P5 true cross-tenant deny incl. the `p5_ambiguous` hard-block, P6 directional partition, own-scope private, tenant discriminator differs, no shared-cache signal, canary survives a fresh URL, synthetic-only bytes); any miss writes nothing. `demonstrated_severity` is a **HARD MEDIUM** by construction, enforced at four layers incl. the merged #117 per-tool ceiling. The row passes the merged **#108 MAC + #111 single-surface + #117 ceiling** gates in a record→freeze→3×-verify round-trip.

## Inert by construction (two independent axes)

1. **Provenance gate:** refuses to sign unless every resolved auth profile carries `synthetic:true` + `email_origin:"temp_email"` + `provisioned_via:"bob_auto_signup"` — and nothing stamps those flags until PR-PROV (`summarizeAuthProfile` doesn't even surface them).
2. **Dispatcher dormancy:** the MCP dispatcher calls the handler with one arg, so the `provision`/`fetch_fn` injection defaults to `null` → `blocked_by_design:object_not_self_provisioned` before any probe.

The only positive (write_row) path is structurally dormant on the merged transport. Live self-provisioning + verify re-hash-only routing land in **PR-D**.

## AC-2 binding — note for reviewers

`path_template` is bound to the surface's **recorded** endpoint via `resolveBaselineFromSurface` (the same binding the read-only confirmer uses), so an off-route template is rejected before any probe; #111 ties the signed proof to the claimed `surface_id`.

This **replaced an http-audit scan-trail check that the pre-commit adversarial review proved was tautological** — it matched the producer's own just-written P2 probe row, so it always passed and bound nothing, and was redaction-degraded on top (`/api/accounts` and `/api/widgets` both redact to `/api/REDACTED`). The replacement is the real binding; the descriptor and code comments no longer claim non-circularity they don't have.

## PR-B-deferred items folded in

`offensive-http-common.js` threads a `toolName` param (default `"bob_http_confirm"` → confirmer stays **byte-identical**; the producer passes its own id) through `assertReadOnlyPath`/`normalizePathTemplate`/`originFromState`, and adds `responseIsSharedCacheable` for the cache-cross-fill exclusion.

## Safety (AC-5, mechanically enforced — never prompt)

Pure-consumer identity model (the producer **never** signs up), server-templated CREATE body, create-time foreign-PII re-read, `piiScan` synthetic-allowlist on P1/P2, no LIST/enumeration. The kinky.nl / reset-totp incident class is structurally excluded.

## Honest residuals (stated in code + descriptor, NOT closed here)

- **in-process-key:** a same-UID `node -e` can read the 0600 signing key and hand-MAC a row around every gate — the same boundary #108/#110 ship with; closed only by the deferred offensive-SANDBOX (UID/container) PR.
- **GAP B (AC-2 topology):** `surface.endpoints[]` is agent-authorable via `bob_append_frontier_event`, so AC-2 proves label↔membership consistency + the single-endpoint/single-host hard-scope, **not** ground-truth topology; closing it needs frontier-event provenance (deferred, design §13).

## Tests

New `test/offensive-idor-producer.test.js` (35 seeded tests): the full AC-2/AC-3/AC-5/AC-6 matrix + the record→freeze→verify round-trip, all with an injected `fetch_fn` carrying **real per-viewer body variance** so a re-introduced whole-body `==` would fail the suite. No live target, no live signup. Registered in the manifest.

## Process / gates

Built by an opus agent from the on-disk contract, then driven through a 4-auditor adversarial soundness verification (AC-2 non-circularity, 25 mint conditions, witness + row + gates + inert, byte-identity/collateral/PII) — verdict **GO**, and the one substantive finding (the AC-2 tautology) is fixed in this PR.

Green: `check:syntax`, generators regenerated + `--check` clean, `check:authority-inventory` (`bob_http_idor_confirm` is `evaluator-web` **only** — no read-only/verifier/evidence inheritance), `test:mcp` 2544/0, `test:prompts` 0-fail. `test:package` fails **locally only** on the pre-existing untracked-`docs/*.md` pack-size artifact (clean CI checkout passes).

@codex review
@coderabbitai review

## Release Notes

* **New Features**
  * Added the new MCP web tool `bob_http_idor_confirm` for deterministic cross-identity IDOR confirmation with signed evidence output.
* **Bug Fixes**
  * Strengthened URL/path and identifier safety checks, and tightened shared-cache/ambiguous-origin detection to reduce false/unsafe confirmations.
* **Tests**
  * Added comprehensive oracle tests and updated installer/tool registry, tool manifests, and evaluator web tool budget assertions.
* **Documentation**
  * Refreshed authority inventory counts and added the new tool entry.
* **Chores**
  * Updated tool allowlists/permissions and authority-class mappings.


vmihalis:
## Review response (commit `1d79e43`)

Thanks — full triage against the merge gate. Producer stays INERT throughout; no finding was reachable in production, but I fixed the ones that matter when the live arm lands.

### ✅ Fixed
- **AC-5 piiScan prefix hole (Codex HIGH):** `piiScan` no longer admits any `eval_*`-prefixed email — it EXACT-matches only the actual provisioned synthetic mailboxes. A prefix pass would have admitted a real `eval_uator@victim.com` and defeated the synthetic-only guarantee once armed. Removed `SYNTHETIC_EMAIL_PREFIX`; the unit test now asserts the prefix imposter is flagged.
- **P2′ scope (Claude):** the cache-bust P2′ URL now passes `assertSafeRequestUrl` before dispatch.
- **`replay_context` false contract (both critics):** removed the unwired `replay_context` input + its lease-governance comment from the descriptor; PR-D adds the real `verification-replay-safety` routing.
- **`originFromState` threading (CodeRabbit Major):** `resolveBaselineFromSurface` now threads `toolName` to `originFromState` (default `"bob_http_confirm"` → confirmer byte-identical; producer passes `TOOL_ID`). My AC-2 fix routed the producer through this helper, so the attribution now matters — good catch.
- **`responseIsSharedCacheable` (CodeRabbit nit):** `no-store`/`private` now take precedence over `public`/`s-maxage` (avoids a false-positive cache rejection).
- **inline `require("./paths.js")` (CodeRabbit nit):** hoisted `sessionsRoot`/`assertSafeDomain` to the top-level import.
- **descriptor honesty (brutalist `global_preapproval` / autoApprove + dead-tool-surface):** the description now leads with **"CURRENTLY INERT — do not call yet"** (every call returns `blocked_by_design` pre-PR-PROV), so agents don't waste turns; the `global_preapproval` comment no longer implies an active per-call checkpoint.
- **P4/P4id comments (CodeRabbit nit / brutalist):** clarified — both are anon reads of `O_B` by its real id (P4id is an anon consistency re-read).

### ⚠️ Declined — with reason
- **`VERB_LIKE_TOKEN_RE` message lacks `toolName` (CodeRabbit nit):** kept intentionally tool-agnostic. This message is byte-identical to the original `bob_http_confirm` confirmer; adding `toolName` would diverge the confirmer's rejection string (PR-B byte-identity contract) for no functional gain. The message names no tool by design.

### ⏭️ Deferred to PR-D — with reason
- **`global_preapproval:false` per-call-checkpoint enforcement:** its `authority_class` matches `bob_http_confirm` and the harness allow-list includes all tools (generator-consistent). A distinct operator checkpoint is only meaningful when the tool can actually execute/CREATE objects — i.e. the live arm. PR-D wires it.
- **P4/P4id probe-redundancy semantics:** since `objBId` is the one real server-assigned id, the two anon probes are genuinely identical (redundant, not unsound — both must 401/403). A distinct private-vs-unlisted-by-id discriminator is an oracle refinement best validated against a live target in PR-D.

The producer is doubly inert (provenance gate + `provision=null` dispatcher dormancy), the merged #108/#111/#117 gates pass a record→freeze→3×-verify round-trip, and AC-2 now binds `path_template` to the surface's recorded endpoint via `resolveBaselineFromSurface` (replacing the tautological scan-trail check round 0 flagged). Gates green: offensive family 99/0, `test:mcp` 2544/0, `test:prompts` 181/0.

@codex review
@coderabbitai review

vmihalis:
## Round-2 review response (commit `75966e4`)

Full triage. The producer is INERT in production (provenance gate + `provision=null` dispatcher dormancy), so none of these were reachable today — but each is a latent soundness/safety bug that would matter the moment the live arm (PR-D) executes, so I fixed them now in the foundation rather than carrying them as debt.

### ✅ Fixed (Codex round-2 inline, all in `75966e4`)
- **P1 — Reject mutation-shaped probe paths:** the producer now applies `assertReadOnlyPath` to the bound baseline **and** every dispatched probe target, mirroring the sibling read-only confirmer (`offensive-confirmer.js:79,88`). `normalizePathTemplate` already forces `{id}` to be the **final** segment, but a verb-NAMED collection *before* the id (`/api/reset/{id}`, `/delete/{id}`) still resolved through `resolveBaselineFromSurface` — now rejected before any probe.
- **P2 — Filter profile metadata out of probe headers:** `resolveIdentity` built outbound headers from the raw profile via `buildHeaderProfile`, which `Object.assign`s its first arg verbatim. Post-PR-PROV profiles carry Bob-local fields (`synthetic`/`email_origin`/`provisioned_via`/`email`, plus `credentials`/storage/expiry), so the full profile would have leaked provenance + the synthetic mailbox to the target. Added `PROFILE_METADATA_KEYS` and strip them before building headers.
- **P1 — Scan the cache-bust proof body for PII:** the synthetic-only gate scanned P1+P2 but not P2′, which is another successful A→B proof body whose canary is required before signing (#16). A fresh-URL fetch can return an expanded payload the cached P2 did not, so P2′ now clears the same `piiScan`.
- **P1 — Reject canary-bearing deny bodies:** a P5 401/403 whose body still leaks B's canary is **not** a true third-identity deny — C actually received B's object in the error envelope. The deny path now scans the body the same way P4/P4id are scanned → `blocked_by_design:p5_canary_in_deny_body`.
- **P1 — Prove P3 is A's object before using its tenant:** P3 supplies identity A's tenant discriminator (#14), but the code only checked it did *not* carry B's canary. A stray/error body with an owning-scope key could have made the cross-tenant distinctness unsound. P3 must now be a 200 resource reflecting A's **own** canary → `blocked_by_design:p3_not_identity_a_object`.
- **P2 — Check capture directory symlinks before creating children:** `resolveCaptureDirSecure` now creates + verifies the session dir is not a symlink **before** creating `offensive-runs/` under it, so a recursive `mkdir` never follows a symlinked session dir to the link target. The post-mkdir realpath check is kept as the second line. (No dedicated FS-symlink unit test — consistent with the sibling `appendSignedRowHardened` guards, which the realpath checks already enforce.)

Adds 5 regression tests (verb-prefixed-endpoint rejection, header-metadata non-leak, P2′ foreign-PII block, P5 canary-in-deny block, P3-not-A block).

### ✅ Already fixed last round (brutalist threads, now stale)
- **🟠 high — PII allowlist admits any `eval_*`-prefixed email:** fixed in `1d79e43` (exact-match only against provisioned synthetic mailboxes; `SYNTHETIC_EMAIL_PREFIX` removed; test asserts the prefix imposter is now flagged). Thread is outdated.

### ⏭️ Deferred to PR-D — with reason (brutalist threads)
- **🟡 medium — `global_preapproval:false` contradicted by the autoApprove list:** the tool's `authority_class` matches `bob_http_confirm` and the harness allow-list includes *all* tools (generator-consistent via `permissionsForAllTools()`). A distinct per-call operator checkpoint is only meaningful once the tool can actually execute/CREATE objects — i.e. the live arm. PR-D wires the checkpoint.
- **🟡 medium — P4 and P4id are byte-identical probes:** intentional at HEAD. `objBId` is the one real server-assigned id; there is no separate "synthetic id" axis (as in #110), so the two anon reads are a redundant consistency re-read — both must 401/403. A distinct private-vs-unlisted-by-id discriminator is an oracle refinement best validated against a live target in PR-D. Comments clarified in `1d79e43`.

### ℹ️ By design — with reason (brutalist thread)
- **🟡 medium — tool registered but can never produce a signed row via the MCP path today:** this is the intended INERT state of the foundation-first build sequence (PR-C lands the contract; PR-PROV stamps provenance; PR-D arms the live path). The descriptor now leads with **"CURRENTLY INERT — do not call yet"** (every call returns `blocked_by_design` pre-PR-PROV), so agents don't waste turns. Not a defect — the dormancy is the safety property.

Gates: offensive-idor-producer **40/0**, `test:mcp` **2549/0** (2 pre-existing skips), `test:prompts` **181/0**, `check:syntax` + `check:authority-inventory` clean. No descriptor/generator surfaces touched this round (confirmer + common-helper byte-identity preserved).

@codex review
@coderabbitai review

vmihalis:
## Round-3 review response (commit `14668a6`)

CI was green on `75966e4`; this round completes the oracle's soundness/safety legs from the deeper review (5 Codex P1 + 3 brutalist medium) **plus one MEDIUM I surfaced with an independent opus adversarial soundness audit** before pushing. Producer remains INERT — these make the live arm (PR-D) build on a sound contract.

### ✅ Fixed — Codex round-3 inline (P1)
- **Full-body canary scan:** new `bodyLeaksCanary()` scans the **entire** response body (not the 256 KB `bodyTextOf` cap) for the P4/P4id anon leak, the P5 deny leak, and the P6 deny leak — a canary hidden past the object cap in a large error envelope is no longer missed.
- **P2′ truncation + size gates:** the cache-bust P2′ proof body now clears the `#1` truncation and `#3` size gates like P1/P2.
- **P6 deny body leaks A's canary:** a B→O_A 401/403/404 whose body still carries A's canary means the partition is false → `p6_canary_in_deny_body` (symmetric to the P5 deny scan; + a P6 truncation guard).
- **Prove identity C authenticates:** new **P7 probe** — C must read its OWN object O_C (200 + canary_c) before its 401/403 on O_B counts as an authenticated-but-different-identity deny. A stale/invalid C is otherwise indistinguishable from anon → `identity_c_not_authenticated`.
- **Reject unsafe self-provisioned ids:** object_a/b/c must pass `capturedIdSegmentIsSafe` (no path separator / action punctuation) — `encodeURIComponent` only encodes, it does not reject → `object_id_unsafe_segment`.

### ✅ Fixed — brutalist round-3 (medium)
- **`oAUrl` scope validation (2 critics):** every dynamically built probe target now passes `assertSafeRequestUrl` (was only oBUrl/p2primeUrl); added for oAUrl + the new oCUrl.
- **`buildSignedOffensiveRunRow` exported:** removed it + `assertSingleEndpointSingleHost` from `module.exports`. The row builder signs **without** the mint conditions, so it must not be an internal-caller gate-bypass path.
- **`responseIsSharedCacheable` comment/code mismatch:** the no-store/private suppression is now scoped to the speculative public/s-maxage branch only; definitive `Age>0` / explicit HIT evidence fires regardless. Behavior preserved, comment now matches code.

### ✅ Fixed — adversarial-audit finding (MEDIUM, surfaced + fixed pre-push)
An opus 3-skeptic audit (oracle-bypass / false-negative / data-leak) found that `#15`+`#16` excluded cache cross-fill by the **absence** of a recognized shared-cache fingerprint — so a query-string-ignoring shared cache that stored a `private`/`no-store` body with `Age:0` and no HIT label could serve B's body to A for **both** P2 and the `?_bobcb` P2′ (same path → same cache key) and mint a row mislabeling a cache cross-fill as an origin IDOR. New **`#15b` `cacheInPathWithoutProvenMiss()`** requires an **affirmative-origin** signal: fail closed when a cache is detectably in the path (Age/Via/X-Cache present) without a labeled MISS → `cannot_prove_origin_read_through_cache`. A direct origin read (no cache header) and a labeled-MISS CDN both still confirm (no false-negative on the common case). **Honest residual** (documented in the file header): a truly fingerprint-less query-ignoring cache is indistinguishable from origin — the cross-tenant read is still real, same MEDIUM impact — closed only by a live path-segment cache-buster in PR-D (a query-key buster can't defeat a path-keyed cache, and `{id}` must stay the final segment). The audit's other two lenses (false-negative regression, data-leak) returned SOUND.

Adds 7 regression tests (P6-canary, C-not-authenticated, unsafe-id, P2′-truncated, P2′-oversized, cache-cross-fill-block, labeled-MISS-positive).

Gates: offensive-idor-producer **47/0**, `test:mcp` **2556/0** (2 pre-existing skips), `test:prompts` **181/0**, `check:syntax` + `check:authority-inventory` clean. Confirmer + common-helper byte-identity preserved.

@codex review
@coderabbitai review

vmihalis:
## Round-4 review response (commit `32b01fd`)

CI was green on `14668a6`. Round-4 findings are all refinements of the round-3 soundness legs (the oracle is converging); producer remains INERT.

### ✅ Fixed — Codex round-4 inline
- **(P2) standard cache headers:** `cacheInPathWithoutProvenMiss` + `responseIsSharedCacheable` now also read the RFC 9211 `Cache-Status` (its `fwd=miss` is an affirmative MISS; `hit` is a definitive hazard) and treat `Cache-Status` / `CDN-Cache-Control` presence as a shared-cache-in-path signal. A CDN reporting only the standardized headers no longer slips past `#15`/`#15b`.
- **(P1) P2/P2′ must be resource responses:** new `#7b` requires the cross-tenant reads to be genuine 2xx resource bodies — a 401/403 error envelope that merely echoes B's canary at the field path is rejected (`p2_not_resource_shaped`) rather than counted as "A read B's object".
- **(P2) reject truncated anon P4/P4id:** the anon bodies prove the unauthenticated read does *not* leak the canary; a truncated body could hide a canary past the fetch cap → `p4_body_truncated` (symmetric with the P5/P6/P2′ gates).
- **(P1) prove identity C is *authenticated*, not just that O_C is readable:** the round-3 P7 leg passed if C could read O_C — but a **public** O_C defeats it (a stale C reads it too). New anon **P8** probe requires O_C to be access-controlled: anon is denied (`o_c_not_access_controlled`) and doesn't leak canary_c (`o_c_public_canary_leaked`); only then does C's P7 success prove authentication.

### ⏭️ Deferred — with reason (logged)
- **(brutalist) `global_preapproval:false` vs the autoApprove list** — re-raised each round. The per-call operator checkpoint is only meaningful once the tool can execute/CREATE objects (the live arm); PR-D wires it. (Same disposition as rounds 1–3.)
- **(Codex P1) dotted-suffix endpoint binding in `pathTemplateMatchesEndpoint`** — a **pre-existing, documented residual in the SHARED helper** (the merged #110 read-only confirmer; see the `capturedIdSegmentIsSafe` comment). It is **benign for the producer**: the probe targets the producer's OWN self-provisioned object id on the recorded `{id}` route family (not the recorded endpoint's literal path), and #111 + the AC-2 single-endpoint binding tie the signed row to the routed surface — so it cannot escape to a different asset, and the canary oracle prevents fabrication. Changing the shared matcher alters the merged confirmer's reviewed binding (PR-B byte-identity), so it belongs in a dedicated shared-helper PR that re-reviews the confirmer, not this PR-C round.

Adds 6 regression tests (P2-error-envelope, P4-truncated, O_C-public, O_C-not-gated, Cache-Status-MISS-positive, Cache-Status-ambiguous-block).

Gates: offensive-idor-producer **53/0**, `test:mcp` **2562/0** (2 pre-existing skips), `test:prompts` **181/0**, `check:syntax` + `check:authority-inventory` clean. Confirmer + common-helper byte-identity preserved.

@codex review
@coderabbitai review

vmihalis:
## Round-5 review response (commit `8a394ad`)

CI was green on `32b01fd`. This round corrects three precision bugs in the round-4 cache logic + a dot-segment id gap; two P2s are declined with reasons. Producer remains INERT.

### ✅ Fixed
- **(brutalist) `s-maxage=0` false-negative:** `responseIsSharedCacheable` matched the bare `s-maxage` token, so `s-maxage=0` — which forces shared caches to **revalidate** every time (not a cross-principal hazard) — was wrongly flagged, false-negativing CDN-revalidated reads. Now only a **positive** s-maxage counts.
- **(brutalist) `Via` over-blocks:** `Via` is a general hop-by-hop header set by **any** proxy/gateway, not just caches, so treating it as shared-cache-in-path over-blocked findings behind a non-caching proxy. Removed `Via` from the cache-presence signals (Age/X-Cache/CF-Cache-Status/Cache-Status/CDN-Cache-Control remain).
- **(Codex P1) `UPDATING`/`REVALIDATED` are not proven MISSes:** the proven-MISS regex included `updating`/`revalidated` (and cross-CDN-ambiguous `expired`), but those serve a **stale cached body** while refreshing — the body came from cache, not origin. A cross-fill reporting `CF-Cache-Status: UPDATING` would have been trusted as origin → false positive. Now only the unambiguous origin statuses (`miss`/`dynamic`/`bypass`) count; everything else fails closed. *(Good catch — this was a soundness bug I introduced in round 4.)*
- **(Codex P2) dot-segment object ids:** `capturedIdSegmentIsSafe` accepted `.`/`..` (relative path traversals — `/api/items/..` resolves to the parent), never a real resource id. Now rejected (literal + percent-encoded). Zero-false-positive shared-helper hardening; benefits the confirmer too.

### ⚠️ Declined — with reason
- **(Codex P2) "reject 2xx error envelopes as P2 resources":** the **canary-at-discovered-field-path** (`#8`) is the categorical witness — a 200 carrying B's canary at the *exact* nested path discovered from B's own owner readback **is** a cross-tenant data read regardless of an accompanying `success`/`error` flag (a server returning B's object inside a `success:false` wrapper is still leaking it). Rejecting it would introduce **false negatives**. `#7b` already excludes pure error/status envelopes and soft-404s, and canary-at-path cannot be satisfied without the server having served B's object structure.
- **(Codex P2) "decode JSON before canary leak checks":** the canary is a validated pure-hex 256-bit nonce (`/^[0-9a-f]{64}$/`) with **no distinct JSON-escaped form** (hex never escapes), so the raw-byte scan over the decompressed body is exact. A substring scan is intentionally the **broadest** leak detector — a parsed-leaf walk would be *narrower* and could miss a canary echoed outside the expected leaf.

### ⏭️ Deferred (logged)
- **(brutalist) `global_preapproval` per-call checkpoint** — re-raised each round; only meaningful at the live arm (PR-D).

Adds 4 regression tests (s-maxage=0 positive, Via positive, CF `UPDATING` block, `..` id block).

Gates: offensive-idor-producer **57/0**, `test:mcp` **2566/0** (2 pre-existing skips), `test:prompts` **181/0**, `check:syntax` + `check:authority-inventory` clean. Confirmer byte-identity preserved (offensive-confirmer suite green despite the shared `capturedIdSegmentIsSafe` hardening).

@codex review
@coderabbitai review

vmihalis:
## Round-6 review response (commit `795592b`)

CI was green on `8a394ad`. All 6 Codex findings (1 P1 + 5 P2) fixed — deep precision hardening of the proof-body synthetic-only guarantee and the cache-origin signal. Producer remains INERT.

### ✅ Fixed
- **(P2) dup-key PII evasion:** `piiScan` ran on the parsed body, but `JSON.parse` keeps only the last value of a duplicate key, so foreign PII in a shadowed duplicate evaded it. `#17` now scans **both** the parsed body and the raw response text for P1/P2/P2′.
- **(P1) injected secrets in the proof body:** the signed stdout capture is the full P2 body, but only PII was screened. New `#17b` `secretShapesIn()` screens for high-confidence credential shapes (JWT / AWS key / PEM / GitHub+Slack tokens) → `non_synthetic_secret_in_response`. The 256-bit hex canary matches none of them (no false-positive).
- **(P2) `\u`-escaped canary in deny bodies:** `bodyLeaksCanary` scanned only raw bytes, so a 401/403 body that JSON-`\u`-escapes the canary (`b...`) decoded to it but evaded the scan. It now also scans the JSON-decoded form. **This reverses my round-5 decline of the same class** — the specific `\u`-escape mechanism is a real evasion I dismissed too quickly; good re-raise.
- **(P2) underscore-delimited cache tokens:** the `\bhit\b`/`\bmiss\b` boundaries missed Varnish/Squid `TCP_MISS`/`TCP_HIT` (`_` is a word char). Both cache helpers normalize `_`→space first.
- **(P2) quoted `s-maxage`:** the value regex now accepts the HTTP-legal quoted form `s-maxage="60"`.
- **(P2) path traversal in templates:** `assertReadOnlyPath` now rejects a `.`/`..` segment, literal **or** percent-encoded at any depth (`/api/%252e%252e/admin`) — `decodePathSegments` resolves to a fixed point, so a server-decodable double-encoded traversal is caught. Shared-helper hardening; zero false-positive; protects the confirmer too.

Adds 6 regression tests (dup-key PII, injected JWT, `\u`-escaped P5 canary, `TCP_MISS` positive, quoted `s-maxage` block, double-encoded traversal block).

Gates: offensive-idor-producer **63/0**, `test:mcp` **2572/0** (2 pre-existing skips), `test:prompts` **181/0**, `check:syntax` + `check:authority-inventory` clean. Confirmer + common-helper byte-identity preserved.

@codex review
@coderabbitai review

vmihalis:
## Round-7 review response (commit `c41b9d0`)

CI was green on `8a394ad`. All 8 threads fixed — **plus 3 findings I surfaced with a focused opus adversarial audit** (`wf_b61d6918-693`) on the cache-signal / URL / escape logic before pushing, to converge the cycle rather than dribble fixes over more rounds. Producer remains INERT.

### ✅ Fixed — review threads
- **(brutalist) `X-Served-By` + `CDN-Cache-Control` over-block:** removed both from the cache-in-path signal — `X-Served-By` is a Fastly/Heroku *node id* on every response, `CDN-Cache-Control` is an origin-authored directive; neither proves a cache handled the read. Cache-presence is now only genuinely cache-generated headers.
- **(Codex P1) network-path reference:** `normalizePathTemplate` now rejects a leading `//` (`//host/{id}` passes the scheme check but `new URL` resolves it to a *different host*).
- **(Codex P2 ×2) escaped+shadowed evasion:** a canary/PII that is *both* `\u`-escaped *and* in a shadowed duplicate key evaded the raw scan (escape) and `JSON.parse` (dup key). Unified on a raw-text `\u`-decode (dup keys preserved); `bodyLeaksCanary` + the `#17` PII + `#17b` secret scans all use parsed + raw + `\u`-decoded.
- **(Codex P2) full P3 canary scan:** `#9` now uses the full-body `bodyLeaksCanary(P3, canary_b)`.
- **(Codex P2) underscore tokens + quoted/spaced `s-maxage`:** normalize `_` so `TCP_MISS`/`TCP_HIT` are seen; accept `s-maxage = "60"`.
- **(brutalist) `global_preapproval` sequencing:** clarified that the dispatcher's `provision=null` dormancy is an **independent structural** guarantee — PR-PROV (provenance only) cannot activate the tool; PR-D wires self-provisioning + the operator checkpoint together.

### ✅ Fixed — focused-audit findings (surfaced + fixed pre-push)
- **HIGH (cache false-positive):** the cache-status reader set omitted the canonical reverse-proxy/CDN HIT headers — **nginx `X-Cache-Status`, `X-Proxy-Cache`, Google `x-goog-cache-status`, Akamai `X-Cache-Remote`, EdgeCast `X-EC-Cache`, Varnish `X-VCache`**. A path-keyed shared cache cross-filling B's body to A labeled with only one of those (no `Age`/`X-Cache`/`CF-Cache-Status`) passed **both** `#15` and `#15b` → bogus origin-IDOR row. New shared `CACHE_STATUS_HEADERS` set drives both the HIT and MISS/presence checks. This is *not* the conceded zero-header residual — the cache is detectably named.
- **MEDIUM (PII evasion):** spaced/dashed PANs (`4111 1111 1111 1111`) evaded `piiScan` (the upstream card matcher is contiguous-only). `piiScan` now also scans a separator-joined normalized copy → Luhn-checkable.
- **LOW (PII evasion):** a trailing-dot FQDN email (`victim@corp.com.`) evaded the email regex; the same normalized copy strips a boundary trailing dot.

The audit verified `responseIsSharedCacheable` / `cacheInPathWithoutProvenMiss` / `piiScan` are **producer-only** (not used by the merged confirmer), so these are byte-safe for `bob_http_confirm`. Its URL/escape lens found no further confirmed issues beyond the `//` and full-P3 items already in the review.

Adds 12 regression tests. Gates: offensive-idor-producer **75/0**, `test:mcp` **2584/0** (2 pre-existing skips), `test:prompts` **181/0**, `check:syntax` + `check:authority-inventory` clean. Confirmer byte-identity preserved.

@codex review
@coderabbitai review

vmihalis:
## Round-8 review response (commit `53e6969`)

CI was green on `c41b9d0`. 4 Codex P2 fixed; 1 brutalist deferred. All refinements of round-7 additions; producer remains INERT.

### ✅ Fixed
- **single-host over-restricts subdomain surfaces:** `assertSingleEndpointSingleHost` seeded the bare session origin, so a legit single endpoint on an in-scope **subdomain** (session `example.com`, endpoint `api.example.com`) falsely counted as two hosts and was rejected. Now counts the endpoint-**resolved** origins (absolute → own, relative → session) + `surface.hosts`, without the bare seed. Genuine multi-host still rejected. (Dropped the now-unused `resolveSurfaceOrigins` import.)
- **`X-Cache-Hits` numeric hit count:** Varnish/Fastly report `X-Cache-Hits: 1` (a count, not a `hit` token) and may omit `Age`/`X-Cache`. `responseIsSharedCacheable` now treats a positive count as a definitive hazard; `cacheInPathWithoutProvenMiss` treats `0` as a proven miss and any presence as cache-in-path.
- **multi-character PAN separators:** the de-group joined only a single separator → `4111  1111 - 1111.1111` evaded. Now joins a run (`[ .-]+`).
- **fine-grained GitHub PAT + more prefixes:** `github_pat_…` didn't match the `ghp_` pattern. Added `github_pat_`, GitLab `glpat-`, Google `AIza`, Stripe `sk_live_`/`rk_live_` (distinctive prefixes; none match the 64-hex canary).

### ⏭️ Deferred (logged)
- **`global_preapproval` per-call checkpoint** — re-raised each round; live arm / PR-D. The dispatcher's `provision=null` dormancy is the independent structural guarantee (PR-PROV alone cannot activate).

Adds 5 regression tests. Gates: offensive-idor-producer **80/0**, `test:mcp` **2589/0** (2 pre-existing skips), `test:prompts` **181/0**, `check:syntax` + `check:authority-inventory` clean. Confirmer byte-identity preserved.

@codex review
@coderabbitai review

vmihalis:
## Round-9 review response (commit `0263fea`)

CI was green on `53e6969`. 6 fixed; 2 deferred; 1 documented. Producer remains INERT.

### ✅ Fixed
- **unified cache hit/miss model:** replaced the ad-hoc cache checks with three shared predicates (`cacheReportsHit` / `cacheReportsMiss` / `cacheDetectablyInPath`). This closes two real bugs: (a) an affirmative MISS (e.g. `X-Cache-Hits: 0`, `CF-Cache-Status: MISS`) now **suppresses the speculative `public`/`s-maxage` hazard**, so a CDN-labeled-MISS public read no longer false-negatives as `cache_shared_response`; (b) `cacheInPathWithoutProvenMiss` **fails closed on any hit signal**, so one cache layer's miss can't override another layer's hit.
- **exact Vary token parsing:** the credential-Vary suppression used a substring test → `Vary: x-authorization-id` wrongly suppressed the hazard. Now parses exact comma-separated tokens (`authorization`/`cookie`/`*`).
- **percent-encoded canary object id:** mint `#19` compared only the raw id; a server-minted id that is a *percent-encoded* canary (`%62%62…` = `bb…`) slipped past but decodes to the canary server-side. Now compares the raw **and** fixed-point percent-decoded form.
- **P3 truncation:** a truncated P3 could hide B's canary past the fetch cap → added a P3 truncation guard (symmetric with P4/P4id/P5/P6) → `p3_body_truncated`.
- **OpenAI-style keys:** added `sk-`/`sk-proj-` to the secret-shape set.

### ⏭️ Deferred (logged)
- **`global_preapproval` (both framings):** the per-call operator checkpoint is a live-arm/PR-D concern; the dispatcher's `provision=null` dormancy is the independent structural guarantee.

### ⚠️ Documented — within the conceded boundary (not a new gap)
- **`{ fetch_fn, provision }` injection seam:** this is **within the already-conceded in-process same-UID boundary** that the offensive-sandbox PR closes — a caller that can pass it can already `require()` the 0600 signing key and hand-MAC a row directly (the exact `#108` assumption; wave-handoff signing makes the identical one). The MCP dispatcher always calls `idorConfirm(args)` with no second arg. Extended the file-header boundary note to state this explicitly.

Adds 7 regression tests. Gates: offensive-idor-producer **87/0**, `test:mcp` **2596/0** (2 pre-existing skips), `test:prompts` **181/0**, `check:syntax` + `check:authority-inventory` clean. Confirmer + common-helper byte-identity preserved.

@codex review
@coderabbitai review

vmihalis:
## Round-10 review response (commit `c0de92d`)

CI was green on `0263fea`. 2 brutalist correctness fixes; 1 deferred. Both simplify the cache-origin discriminator to a more principled model. Producer remains INERT.

### ✅ Fixed
- **speculative Cache-Control directive removed (false-negative):** `responseIsSharedCacheable` flagged a `public`/`s-maxage` response as a hazard *even when no cache was detectably in path*. A bare Cache-Control directive is the **origin's cacheability hint**, not evidence a cache acted — so this false-negatived every legit IDOR on a resource the origin merely marks cacheable. `responseIsSharedCacheable` is now just the **definitive-HIT** signal (Age>0 / explicit HIT token / positive `X-Cache-Hits`); the real residual cross-fill risk (cache detectably in path, no proven miss) stays handled, fail-closed, by `cacheInPathWithoutProvenMiss` (`#15b`). Drops the now-moot s-maxage-value and Vary sub-logic.
- **`X-Cache-Hits: 0` is not a proven miss:** a numeric 0-count is weaker / cross-CDN-ambiguous than an explicit MISS token and could mask a cross-principal hit, so `cacheReportsMiss` now requires an explicit `miss`/`dynamic`/`bypass` token. A response carrying only `X-Cache-Hits: 0` is cache-in-path with no proven miss → `#15b` fails closed.

**Net model:** block on a definitive HIT (`cache_shared_response`); fail closed when a cache is detectably in path without an explicit miss token (`cannot_prove_origin_read_through_cache`); confirm on an explicit miss token **or** when no cache is detectable at all (a directive alone is not cache-presence). The fingerprint-less-cache residual is unchanged (PR-D path-segment cache-buster).

### ⏭️ Deferred (logged)
- **`global_preapproval` cross-PR enforcement** — live-arm/PR-D; the dispatcher's `provision=null` dormancy is the independent structural guarantee.

Reconciled the cache test matrix to the simpler model (a public/s-maxage directive with no cache header now CONFIRMS — was a false-negative block; `X-Cache-Hits: 0` now fails closed; removed the Vary + BWS-s-maxage tests that exercised the removed sub-logic).

Gates: offensive-idor-producer **84/0**, `test:mcp` **2593/0** (2 pre-existing skips), `test:prompts` **181/0**, `check:syntax` + `check:authority-inventory` clean. Confirmer byte-identity preserved.

@codex review
@coderabbitai review

vmihalis:
## Round-11 review response (commit `b214bbb`)

CI was green on `c0de92d`. 5 fixed; 1 declined; 2 deferred. Producer remains INERT. Several of these close entire evasion *classes*, not single cases.

### ✅ Fixed
- **percent-encoded canary/PII/secret (Codex P1/P2):** `bodyLeaksCanary` and the `#17`/`#17b` proof scans now add a **percent-decoded** form (next to raw + `\u`-decoded), so a value hidden as `%xx` in a deny/proof body (survives `JSON.parse`+`canonicalJson` but decodes server-side) is caught.
- **`account_id` is not a tenant (Codex P1):** removed it from `OWNING_SCOPE_KEYS` — it's commonly the record's **own** id, so two records carry different `account_id`s even in the same tenant, which would falsely satisfy the cross-tenant discriminator (`#14`). Remaining keys are unambiguously tenant/owner-level.
- **control-object canary-id disjointness (Codex P1):** the `#19` check now covers `object_a`/`object_c` too (each vs its own canary, raw + percent-decoded) — an O_A/O_C id that is an encoding of its canary would make the P3/P7 legs a trivial id reflection.
- **probe-audit fail-closed (Codex P2):** `auditConfirmRequest` now reports whether the http-audit write persisted; `runProbe` records a swallowed failure and `idorConfirm` fails closed (`#0` → `probe_audit_failed`) rather than minting a durable row for a probe the circuit breaker never saw. The negative-only confirmer ignores the new return (byte-identical).
- **stale comments (brutalist):** fixed the two my round-10 simplification left (`cacheReportsMiss` no longer claims `X-Cache-Hits:0` is a miss; the `cacheInPathWithoutProvenMiss` doc matches `cacheDetectablyInPath` — Via/X-Served-By excluded).

### ⚠️ Declined — with reason
- **`PARTIAL_HIT` matches `\bhit\b` (brutalist):** blocking on a partial cache serve is the **conservative fail-closed** direction — a partial serve *could* be a cross-fill, and avoiding a false-positive signed row outweighs a missed finding. Keeping `PARTIAL_HIT` as a hit is intentional.

### ⏭️ Deferred — with reason
- **collection-vs-instance template binding (Codex P1):** GAP B (agent-influenced surface topology). The producer probes its OWN self-provisioned object on the recorded `{id}` route; the oracle self-validates the route via B's self-read (P0/P1), `#111` ties the row to the surface, and the canary prevents fabrication. Frontier-event provenance is the deferred closure (design §13 / PR-PROV).
- **`global_preapproval` checkpoint:** live-arm/PR-D; the dispatcher `provision=null` dormancy is the independent structural guarantee.

Adds 4 regression tests (percent-encoded P5 canary, percent-encoded PII, percent-encoded control-object id, `account_id`-not-a-scope). `probe_audit_failed` is exercised by the `auditConfirmRequest` return contract + the `#0` check (an FS-permission path not reliably forced in-test).

Gates: offensive-idor-producer **88/0**, `test:mcp` **2597/0** (2 pre-existing skips), `test:prompts` **181/0**, `check:syntax` + `check:authority-inventory` clean. Confirmer byte-identity preserved.

@codex review
@coderabbitai review

vmihalis:
## Round-12 review response (commit `3e98d85`)

CI was green on `b214bbb`. 6 fixed; 1 declined; 1 already-fixed. Producer remains INERT.

### ✅ Fixed
- **malformed-percent decode (P1 — regression I introduced in round 11):** `decodePercentToFixedPoint` decoded the whole string at once, so an unrelated literal `%` (`"100%"`, `"50% off"`) made `decodeURIComponent` throw and returned the body **undecoded** — a percent-encoded canary/PII elsewhere then evaded. Now decodes each valid `%XX` triplet independently (a stray `%` is left alone), looping to a fixed point.
- **method-override headers (P2):** a profile carrying `X-HTTP-Method-Override`/`X-Method-Override` would turn the producer's GET into a server-side mutation. Added `FORBIDDEN_OUTBOUND_HEADERS` and strip them from outbound probe headers.
- **email-as-id / token-as-id (P2):** a self-provisioned id is persisted into the signed row's `target`; every object id (O_A/O_B/O_C) is now screened for non-synthetic PII / credential shapes → `object_id_contains_sensitive_value`.
- **escaped-whitespace PAN (P2):** the de-group now also joins JSON-escaped whitespace (`\n \t \r \f`) and actual `\s` between digits.
- **abort on audit-write failure (P2):** `runProbe` now **throws** on a swallowed http-audit write, so the run aborts immediately (`probe_audit_failed`) instead of firing the remaining un-audited probes (removed the redundant post-loop check).
- **PII match-cap fail-closed (P2):** `piiScan` surfaces a sentinel when a `detectPiiShapes` scan hits `MAX_MATCHES`. *(Defensive — `detectPiiShapes` dedups by value and only 3 mailboxes are allowlisted, so a real PII shape can't actually hide behind the cap, but failing closed on truncation is correct hygiene.)*

### ⚠️ Declined — with reason
- **require an origin read for C's P7 authentication proof:** C reading its **own** object from a cache still proves C is an authenticated principal; the cross-tenant finding's origin-soundness rests on **P2/P2′** (which carry the cache checks). Gating every corroborating leg on origin is scope-creep with no soundness gain.

### ✅ Already fixed (round 11 — duplicate)
- **canary-id disjointness for control objects:** already covers O_A/O_C (`canaryReflectedInId(object_a/object_c, …)`), each raw + percent-decoded.

Adds 4 regression tests. Gates: offensive-idor-producer **92/0**, `test:mcp` **2601/0** (2 pre-existing skips), `test:prompts` **181/0**, `check:syntax` + `check:authority-inventory` clean. Confirmer byte-identity preserved.

@codex review
@coderabbitai review

vmihalis:
## Round-13 review response (commit `e50de8e`)

CI was green on `3e98d85`. All 4 Codex P2 fixed — these close the encoding-completeness and proof-target-screening classes comprehensively (structural closures, not single cases). Producer remains INERT.

### ✅ Fixed
- **unified layered decoder:** a value that *layers* encodings (percent-of-`\u`, `\u`-of-percent, `%2562`) evaded the single-pass scans. New `decodeAllEncodingLayers` composes the `\u` + percent decoders and iterates to a fixed point; `bodyLeaksCanary` and the `#17`/`#17b` proof scans now use it (replacing the separate single-layer decodes).
- **decode object ids before the PII/secret screen:** the round-12 id screen ran on the raw id, so a percent/`\u`-encoded email-as-id evaded it. It now screens the id **and** its fully-decoded form.
- **screen the whole proof target:** the canonical target is **persisted into the signed row**, and its fixed path segments come from the agent-supplied `path_template` — so an email/token in a fixed segment (`/api/users/victim@x.com/{id}`) would leak into the durable proof. The full target (raw + decoded) is now screened → `proof_target_contains_sensitive_value`.
- **drop `Host` + body-framing headers:** a synthetic profile carrying `Host` would override the vhost (routing/cache-key confusion); `Content-Length`/`Transfer-Encoding` enable request smuggling. Added to `FORBIDDEN_OUTBOUND_HEADERS` (safe-fetch derives all of these from the URL/body itself).

Adds 4 regression tests (percent-encoded email-as-id, layered percent-of-`\u` canary, email in a fixed target segment, `Host` strip).

Gates: offensive-idor-producer **96/0**, `test:mcp` **2605/0** (2 pre-existing skips), `test:prompts` **181/0**, `check:syntax` + `check:authority-inventory` clean. Confirmer byte-identity preserved.

@codex review
@coderabbitai review

PR #129, opened 2026-06-18
feat(offense): bob_nuclei_scan — DETECTION-only nuclei lead generator (PR7, lights up the container runner)

vmihalis:
## PR7 — `bob_nuclei_scan`: DETECTION-only nuclei lead generator (the runner's first consumer)

Lights up the **dormant offensive container runner** (`offensive-runner.js` `runOffensiveTool`, which until now was required by nothing but its own test). All four existing signed-row producers (idor/reflect/xss/oob) are in-process (Bob's own HTTP client / headless browser / OOB collector); **nuclei is an external binary that must run in the hardened wide-open container**, so this is the runner's first real consumer.

Produced via an explore → plan → implement workflow over Bob's real source; the architecture fork (generic factory vs. per-binary producer vs. single dispatch tool) resolved to **one runner-backed producer module per binary**, matching the 4 shipped producers (the `offensive-tools.js`/factory PR5 deliverable was never built and isn't needed).

### What it does — leads, never proof
- `offensive-nuclei-producer.js` builds a **server-controlled flag allowlist** + forced flags and a `classify` oracle that **ALWAYS returns `positive:false`**, so the runner never reaches its sign path.
- `bob_nuclei_scan` is **DELIBERATELY ABSENT** from `OFFENSIVE_TOOL_DEMONSTRATED_CEILING` (claims.js) — belt-and-braces: even a future positive verdict would throw `unknown offensive tool_id` rather than mint a row (the **#110 never-sign-server-noise** discipline — a template match is a heuristic, not a categorical witness).
- nuclei output is a **LEAD** the evaluator triages; an OOB/XSS/IDOR-suspect lead is re-proven through the real signing producers (`bob_oob_poll` / `bob_http_xss_confirm` / `bob_http_idor_confirm`).

### 🔴 OAST forced OFF (`-no-interactsh`) — a structural-soundness fix
The original plan ("force nuclei's interactsh to Bob's sink") is **unsound**: nuclei mints its own interactsh correlation IDs in interactsh's wire scheme, which are incompatible with Bob's OOB token (`oob`+32hex), and Bob's sink deliberately doesn't speak the interactsh protocol. So nuclei's OAST is **hard-disabled** — no correlation is attempted, no false OAST signal can arise. A future signed-OOB path injects a Bob-minted token as a **plain template variable** and confirms via `bob_oob_poll`, never via nuclei's `{{interactsh-url}}`.

### Runner change (additive, fail-closed)
On the **existing** negative-classify path, the runner now surfaces a bounded `detection` summary on `masked_summary` — re-serialized, length-capped (16 KiB), and **secret-scanned** (redaction-before-agent), dropped fail-closed on any trip. **No row is ever written on this path.** Also adds `defaultOffensiveDocker.inspectImage` for the fail-closed image preflight. The existing `offensive-runner.test.js` (37/37) is unchanged — the change is backward-compatible.

### Safety (inherited from the runner, not re-implemented)
Wide-open hardened container; fail-closed flag allowlist; AIM-check scope-gating **every** URL flag value; full-stdout secret-scan backstop; per-session run/infra budgets; the `CONTAINER_RUN` audit row. `target_url` is **also** scope-validated at dispatch (`scope_url_fields`). Granted to **`evaluator-web` only** (narrow grant, asserted by `check:authority-inventory`).

### Scope: nuclei detection only
**IN:** the producer, the descriptor, the runner detection channel + inspect seam, authority/registry/drift wiring, 16 seeded tests.
**DEFERRED (explicit follow-ups):** the signed nuclei-OOB path (token-as-template-var → `bob_oob_poll`; needs the live OOB sink provisioned + a nuclei-matched-endpoint→surface_id resolver); sqlmap detection-only; dalfox canary; ffuf/katana recon. Also deferred (cross-cutting, gates any signed runner path): binding `command_hash` to the real container argv (the runner still signs with a placeholder `method:'GET'/runIdPrefix:'reflect'`, **inert here** because this tool never signs).

### Honest boundary (unchanged by this PR)
The offensive-SANDBOX UID/network separation that makes the HMAC absolutely un-fakeable remains deferred (the #108/#128 in-process-key residual). **DETECTION-only adds NO signed row, so it does not widen that gap.**

### INERT-UNTIL-PROVISIONED
Returns `blocked_by_infra:offensive_image_unavailable` until the operator re-mints the digest-pinned arsenal image to include the nuclei binary (`scripts/build-offensive-image.sh`) — ships dormant, fully seam-tested without docker or a live network.

### Tests / gates
- `test/offensive-nuclei-producer.test.js` — **16 seeded, in-process** (stubbed docker + injected deps, no live network): JSONL summary bounds/caps/skips, no extracted-results/response-body leak, severity/tag normalization, classify always-negative, **ceiling-absent assertion**, forced-OAST-off in the container argv, scope-gate rejection, image-not-pinned/not-present → `blocked_by_infra`, secret-scan redaction, never-writes-a-row, module hygiene (no `child_process`).
- Full suite green in CI: `test:mcp` 2759, `test:prompts` 181, `test:install` 15, `test:package` 20, syntax + authority-inventory clean, `offensive-runner` 37.
- Regenerated surfaces: `EXPECTED_TOOL_NAMES`, install count 157→158, authority inventory, evaluator agent/spawn/skill/settings, evaluator-web tool budget 51→52.

### ⚠️ Reviewer flags
- **pack-size budget raised 3.2 → 3.3 MB** (`test/package.test.js`): the lean (committed-tree) tarball had fallen to ~5 KB of headroom; this PR's source consumed it, and leaving 5 KB would make the next unrelated PR fail CI. The lean pack is ~3.20 MB; 3.3 MB restores ~105 KB margin. (The biggest single packable asset is `docs/hacker-bob-social.png` at 921 KB — a candidate to exclude later if the budget tightens again.)
- Local `npm test` `test:package` will fail **only** because of pre-existing **untracked** design docs in `docs/*.md` (PR6 plan, IDOR design, etc., ~282 KB) that the `files` glob packs from the working tree; they are not in this commit or a clean CI checkout. Verified: committed-tree + this PR's files packs to 3,194,934 < 3,300,000 = PASS.

## Summary by CodeRabbit

- **New Features**
  - Added a new detection-only web scanner tool (`bob_nuclei_scan`) supporting `target_url`/`target_domain` with optional severity and template tag filtering.
  - Added bounded “detection” summaries for negative scan outcomes.

- **Bug Fixes**
  - Improved fail-closed safety around preflight/infra failures and redaction/truncation of scan summaries.

- **Documentation**
  - Updated tool/authority inventories and configuration to reflect the new scanner capability.

- **Tests**
  - Expanded nuclei detection producer and end-to-end coverage; updated tool-surface/manifest and packaging size expectations.

- **Chores**
  - Updated the offensive image build and pack size budgets to include version-pinned Nuclei support.


vmihalis:
@codex review

vmihalis:
@coderabbitai review

vmihalis:
## Round-2 review fixes (commit `877d354`)

Thanks — all findings were legitimate and are addressed. Summary of dispositions:

| Finding | Source | Disposition |
|---|---|---|
| Templates can follow redirects off-scope | Codex P1 | **Fixed** — forced `-disable-redirects` |
| Unbounded full-corpus / invasive templates / budget bypass | Codex P1 | **Fixed** — forced `-rate-limit=50 -concurrency=10 -timeout=10 -exclude-tags=intrusive,dos,fuzz` |
| `matched-at` leaks query/credential values to the agent | Codex P2 / brutalist MED | **Fixed** — redacted via Bob's canonical `redactUrlSensitiveValues` before surfacing |
| Doesn't bind session egress policy | Codex P1 | **Fixed (principled subset)** — fail closed when `block_internal_hosts` is set. The offensive container is operator-locked **wide-open** egress (no proxy/gate, `OFFENSIVE_BOB_BUILD_PLAN §1`), so it cannot route through a session egress proxy by design; rather than silently bypass an internal-blocking policy, it now refuses (`blocked_by_design`), mirroring the in-process producers' refuse-when-can't-honor convention. The default-profile metadata/LAN reach is the documented accepted wide-open residual (§4-holes / §residual). |
| nuclei not staged in the build image → exit 127 | Codex P2 | **Fixed** — added to `scripts/build-offensive-image.sh` (required input + lockfile entry); the Dockerfile `COPY bin/` already globs it |
| `ran:false` for timed-out runs | brutalist MED | **Fixed** — a timeout means the container executed + burned a budget slot → `ran:true` |
| `child_process` hygiene regex too narrow | CodeRabbit Major | **Fixed** — now catches `node:` prefix, ESM `from`, dynamic `import()` |

Tests: `offensive-nuclei-producer.test.js` 16 → 19 (timeout-`ran`, `block_internal_hosts` gate, matched-at redaction). The core invariants are unchanged and still covered: `classify` always `positive:false`, absent from `OFFENSIVE_TOOL_DEMONSTRATED_CEILING`, OAST forced off, never writes a signed row.

@codex review
@coderabbitai review

vmihalis:
## Round-2 Codex findings addressed (commit `f385a03`)

| Finding | Sev | Disposition |
|---|---|---|
| Refuse/bind proxy-backed egress profiles | P1 | **Fixed** — the session-egress fail-closed gate now also refuses a non-default (proxy/VPN) `egress_profile`, not just `block_internal_hosts`. The wide-open container is direct-egress and can't honor either; failing closed avoids leaking the operator's real egress for an attribution-bound session. |
| Stage nuclei templates with the binary | P1 | **Deferred-with-reason** — nuclei needs a templates corpus staged in the image (it forces `-disable-update-check`); staging a pinned corpus is part of the deferred live-arm provisioning (with the OOB sink + Bob-owned domain). Documented in `build-offensive-image.sh`. The scan-error fix below means a no-templates run reports `nuclei_scan_error`, **never** a false "no leads". |
| Force `-omit-raw` for JSONL | P2 | **Fixed** — forced `-omit-raw`; JSONL ships request/response pairs by default which the lead summary never needs and which can carry session/CSRF tokens into the captured stream. |
| Treat non-zero nuclei exits as scanner failures | P2 | **Fixed** — `classify` now reads `exitCode`; a non-zero exit surfaces `reason:"nuclei_scan_error"` + `detection.scan_ok:false/exit_code`, not a false negative-detection. |
| Reject URL secrets before building argv | P2 | **Fixed** — reject `target_url` userinfo and strip any fragment before the argv (avoids a `ps` / `docker inspect` leak that bypasses audit + lead redaction). |

The `egress_profile` branch shares the unit-tested `block_internal_hosts` gate structure; it isn't separately tested because the egress-profiles file is at a fixed repo-relative path (not the temp HOME) and needs proxy-identity binding.

Tests: `offensive-nuclei-producer.test.js` 19 → 22; `test:mcp` 2765/0. Core invariants unchanged (classify always `positive:false`, absent from the ceiling registry, OAST off, never signs).

@codex review
@coderabbitai review

vmihalis:
## Merge decision — round-4 findings declined-with-reason

CI is green on `f385a03` (Bob Diff Review, test 20, test 22, CodeRabbit, brutalist). The round-4 brutalist surfaced **2 MEDIUM** findings (Codex critic), both the same structural point about nuclei templates. Declining both with reason and merging on the operator criterion (no unaddressed *genuine* soundness hole — a detection-only tool signs nothing and has no signed artifact; declined/deferred-with-reason don't block).

**1. `-disable-redirects` doesn't constrain template-defined auxiliary targets** (line 71) — **Declined / deferred.** Correct: nuclei workflow templates can encode follow-up targets/DNS/protocol checks the runner's seed-`-u` AIM-check doesn't see. But the critic itself scopes it: *"the risk is bounded to the in-scope domain's surrounding infrastructure (subdomains, auxiliary ports) … worth documenting that -disable-redirects is not a full substitute … A future hardening step would inject a scope-aware HTTP proxy."* That off-scope reach is the **operator-locked wide-open-container accepted residual** (`OFFENSIVE_BOB_BUILD_PLAN §1/§residual` — the container has no egress gate by design). The hard boundary (scope-aware proxy) is explicitly out of the locked wide-open scope.

**2. `-exclude-tags` is a denylist, not a hard boundary** (line 72) — **Declined / deferred.** Correct: a mis-tagged intrusive template slips a tag denylist. But no flag set is a hard boundary — the **template corpus is the trust boundary**, and it's operator-curated as part of the **deferred live-arm provisioning** (the templates must be staged into the image; see `build-offensive-image.sh` + the deferred-with-reason note). `-exclude-tags` + `-disable-redirects` + rate/concurrency caps are defense-in-depth *on top of* that curated corpus; the caller can further scope via the `tags` input. Forcing an opinionated `-include-tags` allowlist would over-constrain and conflict with operator corpus curation ("harden over measure"). The seed-`-u` AIM-check still bounds the blast radius to the in-scope target.

**Tracked as follow-up** (with the deferred live-arm): stage a curated/pinned nuclei-templates corpus + evaluate a scope-aware container proxy as the hard egress boundary.

### Round summary (all substantive findings addressed)
Across 4 review rounds: redirects-off, rate/concurrency/timeout caps, tag denylist, matched-at redaction, `block_internal_hosts` + non-default-`egress_profile` fail-closed gates, nuclei staged in the build image, `ran`-for-timeout correctness, non-zero-exit → `nuclei_scan_error`, URL userinfo reject + fragment strip, `child_process` regex, two pack-size gates → 3.3 MB. Core invariants test-covered: `classify` always `positive:false`, **absent from `OFFENSIVE_TOOL_DEMONSTRATED_CEILING`**, OAST forced off, never writes a signed row. `test:mcp` 2765/0; producer test 22/0.

Merging (squash).

PR #134, opened 2026-06-19
harden(session): cross-version surface-routes resilience + auth-aware circuit breaker

vmihalis:
## What

Two framework-robustness fixes surfaced by a live web engagement whose session spanned a runtime upgrade. Both are degraded-state handling — **no tool, schema, or registry change** (so no drift/pack/install surfaces move).

## #4 — cross-version `surface-routes.json` brick

**Root cause:** the v2.1 `hunter_agent → evaluator_agent` route-field rename never bumped `SURFACE_ROUTE_VERSION`, so a routes file written by a *prior* runtime passes the version gate but fails per-route validation (`assertNonEmptyString(route.evaluator_agent)`). `readSurfaceRoutesStrict` mapped routes with **no per-route try/catch**, so one stale route aborted the whole read and bricked every consumer that funnels through it (routing, context-budget, offensive confirmers, and the downstream verifier/evidence/grader/reporter agents → the "off-FSM" reflex).

**Fix:**
- **Reader resilience** — quarantine a malformed/duplicate route into `result.malformed_routes[]` + emit `repair_hint` ("re-run `bob_route_surfaces`") instead of throwing the whole read. Top-level shape (unparseable / version mismatch / routes-not-array) still fails hard. The file is fully regenerable.
- **Validate-on-write** — `routeSurfacesInternal` validates every generated route *before* persisting, so a future pack/derivation regression fails fast instead of writing a landmine. (Comment: bump `SURFACE_ROUTE_VERSION` on any breaking route-field change.)
- **`getContextBudget` re-surfaces a malformed *target* surface's reason** — read-all consumers stay resilient, but a targeted budget lookup for a malformed surface must surface the error (the agent can't operate under an undefined budget), not silently fall back to the pack default.

## #5 — auth-aware circuit breaker

**Root cause:** the "breaker" is recomputed from the entire `http-audit.jsonl` each read; `isCircuitBreakerFailure` counted **any** 403 regardless of auth, so ~30 unauthenticated-era 403s (auth challenges) stayed tripped forever and later authenticated 200s couldn't heal them.

**Fix (record-level, uses the `auth_profile` already on each record):**
- Split the classifier: `isHardBlockFailure` (429, timeouts/conn-resets, **and a 403 returned despite an auth profile**) always counts, never reclassified; `isUnauthenticatedForbidden` (403 without auth) is a *candidate* auth-challenge.
- `buildCircuitBreakerSummary` reclassifies an unauth 403 as benign **only on positive, temporally-ordered evidence** — a later authenticated 2xx on the **same host+path**. No such success → still counts → a pure WAF wall stays tripped. Healed challenges are surfaced (not hidden) under new `auth_challenge_hosts` / `auth_challenge_403_count` fields.

**Safety (no evasion):** 429 / timeouts / 403-despite-auth are never reclassified; an authed success *before* the unauth 403s does not heal them; a genuine block stays tripped. No traffic sent, no live block bypassed.

## Tests
- New `test/surface-router.test.js` (7) — quarantine stale/all-stale/duplicate, hard-fail on version mismatch, validate-on-write mechanism, happy-path round-trip.
- New `test/circuit-breaker-auth.test.js` (8) — heal + 5 safety/no-evasion cases + classifier split.
- Updated the two `bob_wave_status` external-shape contract assertions in `mcp-server.test.js` for the additive `auth_challenge_*` fields.
- `test:mcp` 2813/0; `test:prompts` 181/0.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fail fast with clearer errors for targeted context-budget and surface-routing requests when route data is quarantined or malformed.
  * Hardened circuit-breaker healing so only safe unauthenticated `403` challenges can be healed by later authenticated successes; WAF/rate-limit/timeouts and authenticated `403` remain blocked.
* **Improvements**
  * Surface-route persistence and loading now quarantine invalid/duplicate routes, sanitize reasons, and optionally include repair hints; added degraded-route diagnostics.
  * Expanded circuit-breaker reporting with auth-challenge fields and tool attribution.
* **Tests**
  * Added/updated regression suites and expected output shapes for route quarantine and auth-aware circuit-breaker behavior.

### DIFF
--- a/mcp/lib/context-budget.js
+++ b/mcp/lib/context-budget.js
@@ -60,6 +60,15 @@ function getContextBudget(args) {
         briefProfile = route.brief_profile;
         capabilityPackVersion = route.capability_pack_version;
         contextBudget = normalizeContextBudget(route.context_budget, pack);
+      } else if (Array.isArray(routesInfo.malformed_routes)) {
+        // readSurfaceRoutesStrict quarantines a malformed route (e.g. an unsupported context_budget
+        // field, or a stale cross-version schema) so read-all consumers do not brick. But a TARGETED
+        // budget lookup for that exact surface must SURFACE the problem — the agent cannot operate
+        // under an undefined budget — rather than silently fall back to the pack default.
+        const malformed = routesInfo.malformed_routes.find((entry) => entry.surface_id === surfaceId);
+        if (malformed) {
+          throw new Error(`surface_id ${surfaceId} has a malformed route: ${malformed.reason}`);
+        }
       }
     } catch (error) {
       if (!/Missing surface routes JSON:/.test(error.message || String(error))) {
--- a/mcp/lib/context-budget.js
+++ b/mcp/lib/context-budget.js
@@ -52,6 +52,17 @@ function getContextBudget(args) {
 
     try {
       const routesInfo = readSurfaceRoutesStrict(targetDomain);
+      // readSurfaceRoutesStrict quarantines a malformed route (e.g. an unsupported context_budget
+      // field, or a stale cross-version schema) so read-all consumers do not brick. But a TARGETED
+      // budget lookup must SURFACE a quarantined target — checked FIRST, so it fires even when a
+      // duplicate VALID route for the same surface also exists (the file is still corrupt and the
+      // agent cannot trust a silently-chosen budget/pack), not just when no route was found.
+      if (Array.isArray(routesInfo.malformed_routes)) {
+        const malformed = routesInfo.malformed_routes.find((entry) => entry.surface_id === surfaceId);
+        if (malformed) {
+          throw new Error(`surface_id ${surfaceId} has a malformed route: ${malformed.reason}`);
+        }
+      }
       const route = routesInfo.document.routes.find((entry) => entry.surface_id === surfaceId) || null;
       if (route) {
         if (route.capability_pack !== capabilityPack) {
@@ -60,15 +71,6 @@ function getContextBudget(args) {
         briefProfile = route.brief_profile;
         capabilityPackVersion = route.capability_pack_version;
         contextBudget = normalizeContextBudget(route.context_budget, pack);
-      } else if (Array.isArray(routesInfo.malformed_routes)) {
-        // readSurfaceRoutesStrict quarantines a malformed route (e.g. an unsupported context_budget
-        // field, or a stale cross-version schema) so read-all consumers do not brick. But a TARGETED
-        // budget lookup for that exact surface must SURFACE the problem — the agent cannot operate
-        // under an undefined budget — rather than silently fall back to the pack default.
-        const malformed = routesInfo.malformed_routes.find((entry) => entry.surface_id === surfaceId);
-        if (malformed) {
-          throw new Error(`surface_id ${surfaceId} has a malformed route: ${malformed.reason}`);
-        }
       }
     } catch (error) {
       if (!/Missing surface routes JSON:/.test(error.message || String(error))) {
--- a/mcp/lib/http-records.js
+++ b/mcp/lib/http-records.js
@@ -385,7 +385,11 @@ function isCircuitBreakerFailure(record) {
 }
 
 function circuitBreakerRecordKey(record) {
-  const host = record.host || hostnameFromUrl(record.url) || "unknown";
+  // Origin (scheme://host:port), not bare hostname: different services on the same host — a non-default
+  // port, or http vs https — block INDEPENDENTLY, so a success on one must not heal a block on another.
+  let origin = "";
+  try { origin = new URL(record.url).origin; } catch { origin = ""; }
+  if (!origin || origin === "null") origin = record.host || hostnameFromUrl(record.url) || "unknown";
   let path = typeof record.path === "string" && record.path ? record.path : "";
   if (!path) {
     try { path = new URL(record.url).pathname; } catch { path = ""; }
@@ -399,12 +403,11 @@ function circuitBreakerRecordKey(record) {
   // identities can share a profile name (reconfigured/legacy), and a success through one identity
   // must not heal a block reached through another.
   const egressId = (typeof record.egress_profile_identity_hash === "string" && record.egress_profile_identity_hash) ? record.egress_profile_identity_hash : "";
-  // Newline-delimited (not space): a free-form egress_profile name may contain a literal space, which
-  // under a space-join could let two distinct (host, method, path, egress, egressId) tuples collapse to
-  // one key and cross-heal. No component (hostname, HTTP method, parsed pathname, profile name, hex
-  // identity hash) can contain a newline, so this is collision-proof. The key is only ever a Map key
-  // within a single summary computation — never persisted or parsed back — so the format is free to change.
-  return [host, method, path, egress, egressId].join("\n");
+  // JSON-encode the tuple: JSON escapes any interior delimiter/quote/newline in a component (e.g. an
+  // egress_profile name containing a newline), so two distinct tuples can NEVER collapse to one key —
+  // no manual sanitization or "no component contains X" invariant required. The key is only ever a Map
+  // key within a single summary computation (never persisted or parsed back), so the format is free.
+  return JSON.stringify([origin, method, path, egress, egressId]);
 }
 
 function buildCircuitBreakerSummary(records, { surface = null, threshold = CIRCUIT_BREAKER_THRESHOLD } = {}) {
--- a/mcp/lib/http-records.js
+++ b/mcp/lib/http-records.js
@@ -399,7 +399,12 @@ function circuitBreakerRecordKey(record) {
   // identities can share a profile name (reconfigured/legacy), and a success through one identity
   // must not heal a block reached through another.
   const egressId = (typeof record.egress_profile_identity_hash === "string" && record.egress_profile_identity_hash) ? record.egress_profile_identity_hash : "";
-  return `${host} ${method} ${path} ${egress} ${egressId}`;
+  // Newline-delimited (not space): a free-form egress_profile name may contain a literal space, which
+  // under a space-join could let two distinct (host, method, path, egress, egressId) tuples collapse to
+  // one key and cross-heal. No component (hostname, HTTP method, parsed pathname, profile name, hex
+  // identity hash) can contain a newline, so this is collision-proof. The key is only ever a Map key
+  // within a single summary computation — never persisted or parsed back — so the format is free to change.
+  return [host, method, path, egress, egressId].join("\n");
 }
 
 function buildCircuitBreakerSummary(records, { surface = null, threshold = CIRCUIT_BREAKER_THRESHOLD } = {}) {
--- a/mcp/lib/http-records.js
+++ b/mcp/lib/http-records.js
@@ -78,6 +78,11 @@ function normalizeHttpAuditRecord(record, { expectedDomain = null, lineNumber = 
       wave: record.wave == null ? null : parseWaveId(record.wave),
       agent: record.agent == null ? null : parseAgentId(record.agent),
       surface_id: normalizeOptionalText(record.surface_id, "surface_id"),
+      // The originating tool (auditConfirmRequest stamps this for the offensive confirmers; null for
+      // bob_http_scan). Persisted so the circuit breaker can tell a faithful-auth bob_http_scan 403
+      // (auth_profile truthful) from an offensive-confirmer probe (audited WITHOUT auth_profile, so a
+      // genuine authenticated block would otherwise look like a healable unauth challenge).
+      tool: normalizeOptionalText(record.tool, "tool"),
       auth_profile: normalizeOptionalText(record.auth_profile, "auth_profile"),
       checkpoint_mode: normalizeOptionalText(record.checkpoint_mode, "checkpoint_mode"),
       block_internal_hosts: record.block_internal_hosts == null
@@ -385,7 +390,12 @@ function circuitBreakerRecordKey(record) {
   if (!path) {
     try { path = new URL(record.url).pathname; } catch { path = ""; }
   }
-  return `${host} ${path}`;
+  // Method + egress are part of the key so an authenticated GET 2xx cannot "heal" a blocked POST on
+  // the same path (method-specific WAF/CSRF/rate-limit), and a success on one egress profile cannot
+  // heal a per-egress block reached through another.
+  const method = (typeof record.method === "string" && record.method ? record.method : "GET").toUpperCase();
+  const egress = (typeof record.egress_profile === "string" && record.egress_profile) ? record.egress_profile : "default";
+  return `${host} ${method} ${path} ${egress}`;
 }
 
 function buildCircuitBreakerSummary(records, { surface = null, threshold = CIRCUIT_BREAKER_THRESHOLD } = {}) {
@@ -415,9 +425,12 @@ function buildCircuitBreakerSummary(records, { surface = null, threshold = CIRCU
     const host = record.host || hostnameFromUrl(record.url) || "unknown";
 
     // Reclassify an unauthenticated 403 as an auth-challenge ONLY with positive, temporally-ordered
-    // evidence: a later authenticated 2xx on the same host+path. Hard blocks (429, timeouts, 403
-    // despite auth) are mutually exclusive with this and are never reclassified.
-    if (isUnauthenticatedForbidden(record)) {
+    // evidence: a later authenticated 2xx on the same host+path+method+egress. Hard blocks (429,
+    // timeouts, 403 despite auth) are never reclassified. The !record.tool gate restricts healing to
+    // the faithful-auth path (bob_http_scan records auth_profile truthfully); offensive confirmers
+    // audit via auditConfirmRequest WITHOUT an auth_profile but DO stamp tool, so a genuine
+    // AUTHENTICATED block from a confirmer (recorded auth_profile:null) is never healed.
+    if (isUnauthenticatedForbidden(record) && !record.tool) {
       const healedTs = latestAuthedSuccessTs.get(circuitBreakerRecordKey(record));
       const recordTs = Date.parse(record.ts);
       if (healedTs != null && !Number.isNaN(recordTs) && healedTs > recordTs) {
--- a/mcp/lib/http-records.js
+++ b/mcp/lib/http-records.js
@@ -395,7 +395,11 @@ function circuitBreakerRecordKey(record) {
   // heal a per-egress block reached through another.
   const method = (typeof record.method === "string" && record.method ? record.method : "GET").toUpperCase();
   const egress = (typeof record.egress_profile === "string" && record.egress_profile) ? record.egress_profile : "default";
-  return `${host} ${method} ${path} ${egress}`;
+  // The egress IDENTITY hash (not just the profile NAME) is part of the key: two different proxy
+  // identities can share a profile name (reconfigured/legacy), and a success through one identity
+  // must not heal a block reached through another.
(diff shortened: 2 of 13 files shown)


vmihalis:
@codex review
@coderabbitai review

**Reviewer focus — two degraded-state robustness fixes (please be adversarial on the safety properties):**

1. **#5 no-evasion (most important).** The breaker reclassifies an unauthenticated 403 as a benign auth-challenge ONLY when a later authenticated 2xx exists on the SAME host+path (`buildCircuitBreakerSummary`, `circuitBreakerRecordKey`). Is there ANY path where a genuine WAF/rate-limit block gets reclassified and silently un-tripped? Consider: a 403 returned *despite* an auth profile (must stay a hard block via `isHardBlockFailure`), 429/timeouts (must never reclassify), an authed 2xx that precedes the 403s (must not heal), and host/path key collisions. The intent: a real block STAYS a hard stop; only stale unauth challenges that the now-attached session demonstrably passes are dropped from the tally.
2. **#4 reader resilience — does quarantine ever HIDE a problem that should surface?** `readSurfaceRoutesStrict` now quarantines a bad route instead of throwing. Read-all consumers degrade gracefully; targeted lookups must still surface the error — I added that to `getContextBudget` (re-surfaces a malformed *target* surface's reason). Are there OTHER targeted single-surface consumers of `readSurfaceRoutesStrict` that should re-surface a quarantined target rather than silently treat it as absent? (e.g. `findRoutedSurface`.)
3. **#4 validate-on-write** — `routeSurfacesInternal` validates each generated route before `writeFileAtomic`. Confirm it can't reject a legitimately-generated route (false-positive abort) for any current pack/surface type.
4. **Additive shape.** `auth_challenge_hosts` / `auth_challenge_403_count` are new fields on the breaker summary; I updated the `bob_wave_status` contract assertions. Any other shape pin or bounded-budget surface I missed?

`test:mcp` 2813/0; `test:prompts` 181/0. No tools/schemas/registry changed.

vmihalis:
## Review round 1 — resolution (commit `7335d49`)

Thanks — all 7 findings were legitimate and hardened exactly the two safety properties this PR exists for. Fixed:

### ✅ #5 breaker — no-evasion (Codex 3×P1)
- **Method in the heal key.** Key is now `host+method+path+egress` — an authenticated `GET` 2xx can no longer heal a blocked `POST` on the same path.
- **Egress in the heal key.** A success via a different egress profile no longer heals a per-egress block.
- **Auth-ambiguity gate.** `auditConfirmRequest` records `auth_profile:null` for *every* offensive-confirmer probe (it takes no `auth_profile`), so a genuine **authenticated** confirmer 403 looked healable. I now persist `tool` on the audit record (auditConfirmRequest already stamped it; `normalizeHttpAuditRecord` was silently dropping it) and **heal only records with no `tool`** — the faithful-auth `bob_http_scan` workflow. Confirmer 403s are never reclassified. 429/timeout/403-despite-auth hard blocks unchanged; a real WAF block still stays a hard stop.

### ✅ #4 quarantine — actionability (Codex 3×P2)
- **`bob_read_surface_routes` now surfaces `malformed_routes` + `repair_hint`** — without this the all-stale recovery path was silent (`surface_count:0`, no signal to re-route). This is the fix that makes the whole recovery flow actually reach the orchestrator.
- **`findRoutedSurface` re-surfaces a quarantined target** ("…has a malformed route (re-run `bob_route_surfaces`)") instead of a misleading "unknown surface_id".
- **`getContextBudget` checks `malformed_routes` FIRST** (even when a valid route is also found), so a duplicate/conflicting corrupt entry surfaces rather than silently picking one budget/pack.

### ✅ CodeRabbit (Major) — path-leak
`readSurfaceRoutesStrict` sanitizes a quarantine `reason` to the basename (`sanitizeRouteReason`), so a reason surfaced to a caller never leaks the absolute session path.

### 🔎 Incidental
While fixing the heal key I found a stray **NULL byte** had crept into it (`${host}\0${path}` where a space was intended — harmless self-consistent key, but un-greppable/un-editable). Replaced with a clean space-delimited method+egress key (the fields are space-free → collision-safe). The file's two **pre-existing** NULL jwt-key delimiters are deliberately untouched.

Tests: +3 breaker safety regressions (method/egress/tool-stamped-confirmer must NOT heal); updated the `offensive-http-common` contract test to assert `tool` IS now persisted (intentional reversal — it gates the heal logic, never breaker-visibility). `test:mcp` **2816/0**; `test:prompts` **181/0**.

@coderabbitai review
@codex review

vmihalis:
## Review round 2 — resolution (commit `8e908a0`)

First, the **reframe that sets the severity bar** for the breaker items: the circuit breaker is **advisory safety-feedback**, not a security boundary. The evaluator prompt is explicit — *"Use … circuit_breaker_summary to avoid hammering hosts that are repeatedly returning 403/429/timeouts. This is safety feedback, not permission."* It protects the **operator** (don't hammer a blocking host / get IP-banned), not the target. So an over-heal's blast radius is bounded and **self-correcting**: the agent sends a few more requests to a still-blocking host → fresh 403s → the breaker re-trips. No target is compromised. I fixed the items both bots agree are Major/High; the rest are declined under that framing.

### ✅ Fixed
- **`findRoutedSurface` split-authority (Codex HIGH / CodeRabbit Major).** It checked `malformed_routes` only inside `if (!route)`, so a valid-first + malformed-duplicate for the same `surface_id` let live offensive probes proceed while `getContextBudget` rejected — split authority over one artifact. Now checks `malformed_routes` **before** route selection, mirroring `getContextBudget` exactly. +1 regression test (valid + quarantined duplicate → rejects).
- **Egress IDENTITY in the heal key (Codex P1).** Key used the profile NAME, not `egress_profile_identity_hash`; two identities sharing a name could cross-heal. Key is now `host+method+path+egress+egress_identity`. +1 regression test.
- **Bounded `auth_challenge_hosts` (CodeRabbit Major).** Full scalar `auth_challenge_403_count` kept; the host list is capped to `HTTP_AUDIT_SUMMARY_MAX_ITEMS`.

### ⏸️ Declined-with-reason (advisory breaker; over-heal self-corrects; never a target-compromise)
- **Redacted-path heal collision (Codex P1).** The breaker is host/path-level; a query-value-specific 403 (`?id=a` vs `?id=b`) is an **authz** signal, not a host block, and isn't what the breaker counts. The path redaction is a deliberate **PII-safety** feature — and excluding query-bearing endpoints from healing would re-introduce the exact stale-trip this PR fixes (a far worse, common-case regression than a self-correcting host-level over-heal).
- **Raw `Authorization`/`Cookie` 403 without an `auth_profile` (Codex P1).** The audit record doesn't capture raw-header auth presence (only `auth_profile`); this is unusual usage (the supported path attaches an `auth_profile`). A misclassified heal here is advisory + self-correcting.
- **Metadata-only `auth_profile` (Codex medium).** A name-set/zero-header profile is a narrow misconfiguration; same advisory/self-correcting bound.
- **Pre-upgrade confirmer rows (Codex P1).** Old rows written before `tool` was persisted are transient — a **fresh** session stamps `tool` correctly; the contaminated-session path is exactly the case the operator restarts.
- **Empty-toolId convention (Claude medium).** No confirmer passes an empty `toolId`; `normalizeOptionalText("")→null` only on genuinely-empty input.

The heal closes the **common** case (unauth recon → authenticate → success on the same key) and is documented as best-effort. `test:mcp` **2817/0**; `test:prompts` **181/0**.

@coderabbitai review
@codex review

vmihalis:
## Review round 3 — resolution (commit `37a21ad`)

Both round-3 **brutalist** mediums (posted at 17:58, *after* both bots' round-2 `[approve]`) are fixed. Both harden the exact #4/#5 code this PR touches, so I took them rather than deferring.

### ✅ Fixed
- **Collision-proof circuit-breaker key (Claude/brutalist medium).** `circuitBreakerRecordKey` joined `host method path egress egressId` with plain **spaces** — a literal space in a free-form `egress_profile` name could collapse two distinct tuples to one key and cross-heal a block on a *different* path/method. Now newline-delimited: no component (hostname, HTTP method, parsed pathname, profile name, hex identity hash) can contain a newline, so it's collision-proof. The key is only ever a `Map` key within one summary computation (never persisted/parsed back), so the format is free to change. **+1 regression test** — a success whose key collides *only* under the old space delimiter does **not** heal a different block (`tripped_count` stays 1).
- **Surface validator CODE bugs now fail loud instead of being quarantined (Codex/brutalist medium).** The per-route `try/catch` in `readSurfaceRoutesStrict` quarantined *any* throw as a "stale route", so a future programming regression inside `validateSurfaceRoute` (null deref → `TypeError`, renamed import → `ReferenceError`) would be masked behind a misleading "re-run `bob_route_surfaces`" hint. Every data-validation failure in the validator + its helpers (`assertNonEmptyString`, `getCapabilityPack`, `normalizeContextBudget`) is a **plain `Error`**; a JS error **subclass** can only be a code bug — so I re-throw `TypeError`/`RangeError`/`ReferenceError`/`SyntaxError`/`EvalError`/`URIError` loudly and quarantine only plain `Error`. **+1 test** pins the premise (a refactor that threw a subclass on bad data — which the guard would then wrongly re-throw and re-brick — fails the test).

### Already fixed (re-raised)
- **`findRoutedSurface` split-authority (brutalist `🟠 high`).** Fixed in `8e908a0` and **verified by both bots at round 3** — the malformed-routes check now runs *before* route selection (`offensive-http-common.js:255-260`), mirroring `getContextBudget`, with a regression test. The brutalist inline thread predates that commit; it wasn't auto-outdated because it's posted by a bot, not tied to the moved hunk.

### ⏸️ Still declined-with-reason (round-2 advisory-breaker items, logged above)
Unchanged: the breaker is **advisory safety-feedback**, not a security boundary — an over-heal sends a few more requests to a still-blocking host, gets fresh 403s, and re-trips. Redacted-path heal collision, raw-`Authorization`/`Cookie` 403 without an `auth_profile`, metadata-only `auth_profile`, pre-upgrade confirmer rows, empty-`toolId` convention.

`test:mcp` **2820/0** (+2 new tests, 2 pre-existing skips); `test:prompts` **181/0**.

@coderabbitai review
@codex review

vmihalis:
## Review round 4 — resolution (commit `9678d74`)

Round-4 surfaced one **genuinely new** brutalist medium and one re-raise.

### ✅ Fixed — `5tOG` (new): two unpatched `safeSurfaceRouteMap` callers
This was a real incompleteness of *this PR's own #4 fix*. Making `readSurfaceRoutesStrict` tolerant (quarantine instead of throw) meant `safeSurfaceRouteMap` in `prepare-node.js` + `finalize-node.js` — which build the one-hop graph-context surface-metadata map — iterate only `doc.routes` and never inspect `malformed_routes`, so a fully-quarantined file yields `{}` (indistinguishable from "no routes set up").

Worth noting it's strictly **better** than pre-PR: previously the reader *threw* on any bad route, the `catch` returned `{}`, and **all** routes vanished silently. The residual gap is the *silence*. Fix: emit a `process.stderr.write` diagnostic (the codebase's warning channel — cf. `session-authority` shadow-mode, `offensive-http-common` audit-write-failed) naming the quarantined `surface_id`s + repair hint + retained count, while still returning the best-effort map.

These helpers only **enrich** graph context, so they degrade gracefully — deliberately **not** a throw, which would brick `bob_prepare_node`/`bob_finalize_node` on a partially-stale file (the exact brick #4 prevents). That's the principled difference from `getContextBudget`/`findRoutedSurface`, where a wrong/missing route *is* the answer and must hard-fail. Applied identically to both pre-existing copies.

No bespoke test: both modules export a **tool descriptor** (no clean unit seam for the internal helper); valid-route retention is transitively covered by the new surface-router reader tests (`doc.routes` carries only valid routes); the diagnostic is purely additive, gated on `malformed_routes.length > 0` — **false for every existing happy-path prepare/finalize test (14/14 pass, 0 spurious diagnostics confirmed)**.

### ⏸️ Declined-with-reason — `5tOE` (re-raise): `!record.tool` heal gate
Same item declined in round 2/3, now with both bot framings (Claude: "unenforceable convention"; Codex: "pre-upgrade rows retroactively heal-eligible"). Held under the **advisory-breaker reframe**: an over-heal sends a few more requests to a still-blocking host → fresh 403s → re-trips, so it never compromises a target. The current confirmers *do* stamp `tool` (verified), so the "future tool forgets `toolId`" case is hypothetical; the pre-upgrade-rows case is a continued-session transient that a fresh session (the operator's natural response to a corrupt session) resolves. Adding a source-side guard to an advisory mechanism isn't warranted.

`test:mcp` **2820/0**; `test:prompts` **181/0**.

@coderabbitai review
@codex review

vmihalis:
## Review round 5 — resolution (commit `75c564b`)

This round **permanently closes** the breaker-key concern classes and patches the third (last) consumer of the tolerant reader.

### ✅ Fixed
- **`53Md` — origin, not bare hostname, in the heal key.** Different services on one host (a non-default port, http vs https) block independently. Key now uses the URL **origin** (`scheme://host:port`). A `:443` success no longer heals a `:8443` block. +1 regression test.
- **`53_n` — key is now `JSON.stringify([...])`, not a delimiter-join.** JSON escapes any interior delimiter/quote/newline in a component, so two distinct tuples can **never** collapse to one key — collision-proof *by construction*, retiring the "no component contains a newline" invariant the prior newline-join only asserted in a comment. (The 53_n scenario was real: `normalizeOptionalText` trims only edge whitespace, so an interior `\n` in an `egress_profile` name survived.) +1 regression test (`"p\nq"` vs `"p"` stay distinct). Subsumes the round-3 5gmG fix.
- **`53Mg` — `resolveSurfaceTechniqueRoute` now surfaces a quarantined route.** The third tolerant-reader consumer; it silently fell back to `classifySurfaceCapability` re-derivation for a quarantined surface. Re-derivation is the **safe** result (a current-schema pack), so — like prepare/finalize-node and unlike `getContextBudget`/`findRoutedSurface` — it degrades gracefully rather than bricking; added a stderr diagnostic + repair hint so the corruption isn't silent.

### ⏸️ Declined-with-reason
- **`53_y` (discriminator relies on JS error class).** The surface-router test pins the "data failures are a plain `Error`" contract today. A future *transitive* helper throwing a plain `Error` for a code bug is a maintainability concern about validator **growth**, not a current soundness hole — and the test fails loudly the moment that contract is violated.
- **`53_v` (heal premise: an authed 2xx proves earlier unauth 403s were benign).** The core advisory-breaker design objection, answered by the reframe: the breaker is **operator safety-feedback, not a security boundary**. An over-heal sends a few more requests to a still-blocking host → fresh 403s → re-trips. No target is ever compromised.
- **`53Ma` (don't heal across redacted query values).** Round-2 re-raise: a query-value-specific 403 is an **authz** signal, not a host block (which is what the breaker counts), and the path redaction is a deliberate PII-safety feature.

`test:mcp` **2822/0** (+2 new tests, 2 pre-existing skips); `test:prompts` **181/0**.

@coderabbitai review
@codex review

vmihalis:
## Review round 6 — resolution (commit `0671696`)

Two genuine incompleteness items in this PR's own round-5 fixes — fixed. The rest are re-raises of declined advisory-breaker items.

### ✅ Fixed
- **`6C6k` — trim the stored malformed `surface_id`.** A quarantined route that kept a whitespace-padded id (e.g. `" surface:api "`) **evaded** the `getContextBudget`/`findRoutedSurface` corruption-rejection (the 5RG9 fix), because those match the malformed entry's id against the *trimmed* request id while `validateSurfaceRoute` trims on the valid path. The catch now stores the trimmed id, so padding can't bypass the rejection. +1 regression test.
- **`6C6i`/`6EY5` — warn on a quarantined route regardless of `!route`.** The round-5 53Mg diagnostic was gated on `!route`, so a valid-first + malformed-duplicate file (which the hard gates reject *unconditionally*) proceeded **silently** through `resolveSurfaceTechniqueRoute`. The check now fires whenever the surface is quarantined.

On the Codex "make all three share the same decision" ask: technique selection still **degrades gracefully** (uses the retained valid route, or re-derives) rather than bricking — it feeds `bob_select_technique_packs` + `bob_log_technique_attempt`, advisory selection/logging flows, **not** the authoritative offensive-probe gate. Making it *reject* would brick pack-selection/attempt-logging on a merely-stale route, contradicting #4. The authoritative corruption rejection stays in `getContextBudget`/`findRoutedSurface` downstream; the new warning ensures the operator is informed at selection time too (the Claude critic's actual ask — it rated the design sub-point 🔵 low and accepted the degrade-for-quarantined design).

### ⏸️ Declined-with-reason — advisory breaker (an over-heal sends a few more requests to a still-blocking host → fresh 403s → re-trips; never a target-compromise)
- **`6EY4`/`6C6d` (pre-upgrade confirmer 403s lack `tool` → heal-eligible; tagged "security").** The breaker is **operator safety-feedback, not a security boundary** — this cannot compromise a target. Current confirmers stamp `tool`; pre-upgrade rows are a continued-session transient an operator restart clears. A schema-version field on audit records is disproportionate for an advisory mechanism. (3rd raise; rationale unchanged.)
- **`6C6f` (empty/metadata-only `auth_profile` treated as authed).** A name-set/zero-header profile is a narrow misconfiguration; same advisory/self-correcting bound.

`test:mcp` **2823/0** (+1 new test, 2 pre-existing skips); `test:prompts` **181/0**.

@coderabbitai review
@codex review

vmihalis:
## Review round 7 — resolution + merge decision (head `0671696`)

All four round-7 brutalist threads are re-raises / forward-looking design+maintainability concerns. **None is a new soundness or security hole**, so per the merge gate (CI green + threads resolved + no *unaddressed* genuine soundness/security hole; declines-with-reason don't block) this PR is merging.

### ⏸️ Declined-with-reason
- **`6PQO` + `6PQR` (tagged "security") — the `!record.tool` heal gate** ("untyped `tool` field, no producer contract" / "negative-space trust"). 4th raise of the same family (5tOE → 6EY4 → here). The tag is a mischaracterization: the circuit breaker is **operator safety-feedback, not a target-security boundary** — the evaluator prompt is explicit ("safety feedback, not permission"). Worst case of a wrong heal: the agent sends a few more requests to a still-blocking host → fresh 403s → the breaker re-trips. No target is compromised, no data leaks, no unauthorized action occurs, and it self-corrects. Current confirmers stamp `tool`; pre-upgrade rows are a continued-session transient an operator restart clears. A typed producer contract / schema-version field is disproportionate for an advisory mechanism.
- **`6PQX` (maintainability) — JS error-subclass discriminator as an "invisible contract."** Re-raise of round-5 53_y. The `validateSurfaceRoute signals DATA failures with a PLAIN Error` test pins the contract today and fails loudly the instant a refactor violates it. Maintainability of a tested invariant, not a current soundness hole.
- **`6PQY` (design) — caller-defined malformed-route policy may diverge as new consumers are added.** A legitimate forward-looking design point, but every *current* consumer of `readSurfaceRoutesStrict` is handled: `getContextBudget` + `findRoutedSurface` reject unconditionally (hard gates); `prepare-node` + `finalize-node` + `resolveSurfaceTechniqueRoute` warn-and-degrade (advisory enrichment/selection). A central malformed-routes policy (e.g. a tolerant-mode opt-in on the reader, or a lint flagging new callers) is a reasonable **follow-up**, not a blocker for a PR scoped to #4/#5 robustness.

### Why this merges now
The PR's core is solid and proven across 7 review rounds: **#4** cross-version surface-routes resilience (quarantine-not-brick, validate-on-write, split-authority rejection across all hard-gate consumers, whitespace-evasion-proof, advisory consumers degrade-with-diagnostic) and **#5** auth-aware circuit breaker (real WAF stays tripped, 403-despite-auth stays tripped, 429/timeout never healed, method/egress/egress-identity/port isolation, temporal ordering, JSON-encoded collision-proof key). `test:mcp` **2823/0**, `test:prompts` **181/0**; both bots `[approve]`d at rounds 3–6 and CI is green. The remaining open threads are advisory re-raises and forward-looking design notes, all logged above.

PR #135, opened 2026-06-19
feat(offense): "ask before writing" gate — flag-gated PreToolUse HITL confirm on bob_http_scan writes

vmihalis:
## What

An **"ask before writing"** gate: when enabled, Bob pauses and asks the operator to confirm before sending a target-**mutating** `bob_http_scan` request (POST/PUT/PATCH/DELETE). Read probes — and every other tool, including the IDOR/CORS/XSS confirmers (they fire GETs) — keep running autonomously.

Operator request: *"i just want bob to ask me before writing"* + *"only write-method HTTP, make this a flag."*

## How

- **Mechanism = PreToolUse hook.** A hook is the *only* thing in Claude Code that can return `permissionDecision:"ask"` (the MCP runtime can only *refuse*, not ask-then-proceed). `.claude/hooks/bob-http-write-confirm.js` matches `mcp__hacker-bob__bob_http_scan`, and for a mutating method returns an `ask` decision; it abstains (exit 0, no prompt) on flag-off, a read method, a non-scan tool, or unparseable input.
- **Flag-controlled, inert by default.** Does nothing unless `BOB_HTTP_WRITE_CONFIRM` is truthy in Claude Code's env. Default-off preserves Bob's operator-locked **fully-autonomous** default for all other installs; an operator opts in per-workspace.
- **Keys on the outbound HTTP method**, not a tool's `mutating` descriptor flag (which also covers internal audit-ledger writes) — the ask is about *target* writes.
- **Scope = `bob_http_scan` only** (operator's choice): `bob_auto_signup` already carries the #116 temp-email provenance guard.
- **Propagation:** the matcher lives in `adapters/claude/config.js` `defaultPreToolUseHooks()` (the canonical install source-of-truth) so it merges into every target on install — verified empirically (a seeded target gains the matcher exactly once, pointing at the hook). The hook is added to the `files` allowlist so it's packed.

## Contract refinement (the one judgment call)

Two tests deliberately forbade **any** `mcp__` PreToolUse matcher (`mcp-server.test.js` "Claude settings register only artifact guards…" + `prompt-contracts.test.js`), to keep scope/HTTP **enforcement** out of bypassable hooks — that lives in the MCP runtime. The write-confirm gate is categorically different: a **HITL confirmation** (it *asks*; if the hook is removed Bob just proceeds autonomously — the default), **not** enforcement, so it creates no false security assurance. Both contracts are narrowed to forbid scope/**enforcement** hooks on MCP tools while permitting *exactly* this gate. The scope-guard strip-on-merge behavior is unchanged — the merge test still proves a `scope-guard-mcp.sh` hook is stripped; the matcher now survives only via the canonical write-confirm hook.

## Tests

`test/test-http-write-confirm.py` — 18 cases (inert-by-default; the 4 mutating methods ask; reads/HEAD/OPTIONS pass; scope is scan-only incl. `bob_auto_signup` passes; malformed/missing-method abstain; flag truthiness variants). Wired into `test:hooks`.

Green: `test:mcp` 2823/0, `test:prompts` 181/0, `test:install` 15/0, `test:cli` 23/0, `test:hooks` 75/55/18.
`test:package`: committed-tree `npm pack` = **3.223 MB < 3.3 MB** (CI-clean). A local 3.33 MB reflects pre-existing **untracked** `docs/*.md` design docs inflating `npm pack` — not in this branch's commits; proven by excluding them.

@coderabbitai review
@codex review

## Summary by CodeRabbit

* **New Features**
  * Added an opt-in “ask before write” approval gate for `bob_http_scan` HTTP mutations (POST/PUT/PATCH/DELETE). Read-only requests (GET/HEAD/OPTIONS) proceed without approval, including method-override handling to treat mutating overrides as writes.
* **Tests**
  * Added a dedicated test harness covering opt-in flag semantics, method overrides, and fail-closed behavior; tightened assertions to ensure the gate applies only to the intended MCP matcher/tool.
* **Chores**
  * Shipped and installed the new hook scripts and updated package/hook lists.


vmihalis:
Review round 1 — flag-gated PreToolUse write-confirm gate on `bob_http_scan` + the operator-approved contract refinement (HITL confirm permitted; scope/enforcement hooks still forbidden). Design + the committed-tree pack proof are in the PR body.

@coderabbitai review
@codex review

vmihalis:
## Review round 1 — resolution (commit `bdda45f`)

Both brutalist mediums were on the hook; both are fixed by switching to the codebase's established hook idiom (`session-write-guard.sh` = bash + python heredoc).

### ✅ Fixed
- **`6ocq` (Claude, perf) — node spawned on every `bob_http_scan` even when disabled.** The hook is now **bash**: a pure-bash flag check exits immediately (allow) when `BOB_HTTP_WRITE_CONFIRM` is unset/falsy — **no interpreter spawn**. python runs only when the gate is ON. So the default-off case (every install that doesn't opt in) costs nothing per scan.
- **`6ocp` (Codex, correctness) — fail-open on parse error.** The gate now **fails CLOSED**: when enabled, any `bob_http_scan` call it cannot positively confirm to be a **read** (malformed / missing / unknown method) **asks** rather than silently allowing a possible write. Definitive reads (GET/HEAD/OPTIONS) and out-of-scope tools still pass through.

### Rename ripples (`.js` → `.sh`)
The hook became `bob-http-write-confirm.sh`. Updated everywhere: the `config.js`/`settings.json` command (`node`→`bash`), **`adapters/claude/index.js` `HOOK_FILES` + `EXECUTABLE_HOOKS`** so the installer actually **copies** the hook + `chmod +x` (a gap caught while verifying — the matcher pointed at a file the curated install list didn't ship; now a seeded target has the matcher **and** the executable file), `package.json` `files`, the packed-hook test, and the command-match regexes in install-smoke/mcp-server/prompt-contracts. `test/test-http-write-confirm.py` invokes bash + covers the fail-closed cases (20 cases).

`test:mcp` **2823/0**, `test:prompts` **181/0**, `test:install` **15/0**, `test:cli` **23/0**, `test:hooks` 75/55/**20**. `test:package`: committed-tree pack **3.223 MB < 3.3 MB** (local red = pre-existing untracked `docs/*.md` design docs, not this branch).

@coderabbitai review
@codex review

vmihalis:
## Resolution (commit `cc2babb`) — late round-1 reviews

### ✅ Fixed
- **`6pYI` (Codex P2, PII) — redact the confirmation URL.** The reason printed the raw `bob_http_scan` URL before the MCP handler's redaction, so a write replay of an OAuth callback / signed URL / captured-traffic URL could surface `token`/`code`/`id` from the query or fragment in the prompt (and transcript). The hook now redacts to **origin+path** (drops userinfo/query/fragment) — the codebase's value-blind redaction. Verified: `?code=…&token=…#frag` → `scheme://host/path`, no secret.
- **`6uDm` (CodeRabbit Major, test integrity) — anchor the contract assertions.** The MCP-tool hook-command assertions used a bare substring match that would pass on a chained command (`bash "…sh" && evil`). Both `mcp-server.test.js` assertions (Part A + Part B merge) and the sibling `prompt-contracts.test.js` assertion now use an **anchored `^…$` exact-command match** (a bash invocation of the gate, nothing appended).

### Already fixed
- **`6pYD` (Codex P1) — add the hook to the installer copy list.** Done in `bdda45f`: `adapters/claude/index.js` `HOOK_FILES` + `EXECUTABLE_HOOKS`. Verified a seeded target now gets the matcher **and** the executable hook file.

### ⏸️ Declined-with-reason
- **`6pYL` (Codex P2) — cover wrapper tools that forward writes to httpScan.** The scope is deliberately **`bob_http_scan`-only** (the operator's explicit choice). The one write-forwarder, `bob_auto_signup`, is intentionally out of scope and already carries the #116 temp-email provenance guard; the offensive confirmers fire GETs. A future write-forwarding wrapper would be a deliberate scope extension, not a gap in this PR.

`test:mcp` **2823/0**, `test:prompts` **181/0**, `test:hooks` 75/55/**20**.

@coderabbitai review
@codex review

vmihalis:
## Review round 2 — resolution (commit `327fc6b`)

The hook is now a thin bash flag-check that exec's a python impl reading **stdin** (`bob-http-write-confirm.sh` + `bob-http-write-confirm-impl.py`).

### ✅ Fixed
- **`61pk` (Codex P1) + `63be` (brutalist) — payload in the environment.** It was `export`ed into the python child's env. A large captured request body (≥ `ARG_MAX`) made the `python3` exec abort with *"Argument list too long"* **before any decision** — a non-2 exit, which Claude Code treats as **non-blocking**, so exactly the large mutating writes that need confirmation would slip through **unprompted**. Now the payload is read from **stdin** (no env/argv → no size limit, and out of `/proc/<pid>/environ`). The impl always exits 0 (ask/allow), failing closed on any error.
- **`61pi` (Codex P2) — method-override on a read method.** A `GET`/`HEAD`/`OPTIONS` that smuggles a write via `?_method=DELETE`, `X-HTTP-Method-Override`/`X-Method-Override`/`X-HTTP-Method: <mutating>`, or a `_method=<mutating>` form body now **asks** (httpScan forwards those to `safeFetch`). Mirrors how the offensive read probes treat `_method`/override as mutation-shaped. +6 test cases.

### ⏸️ Declined-with-reason
- **`61pf` (Codex P1) — redact secret-looking PATH segments.** The confirmation prompt is **operator-only and ephemeral** — not a persisted/signed/shared artifact like the offensive-runs ledger that `redactUrlSensitiveValues` protects. origin+path is the **decision context** the operator needs (POST `/api/users` vs POST `/api/admin/wipe`), and the high-risk shared-secret replay vectors (OAuth `?code=`, signed-URL `?token=`) live in query/fragment, **already dropped**. Reimplementing entropy heuristics — or coupling a standalone hook to the runtime redactor — is disproportionate for the operator's own prompt.

Verified a seeded install copies **both** hook files and the gate runs end-to-end (a `?token=` URL renders origin+path, no secret). `test:hooks` write-confirm = **26 cases**; `test:mcp` 2823/0, `test:prompts` 181/0, `test:install` 15/0, `test:cli` 23/0; committed-tree pack 3.224 MB < 3.3 MB.

@coderabbitai review
@codex review

vmihalis:
## Review round 3 — resolution (commit `4a5ebc9`)

### ✅ Fixed
- **`6_xR` (brutalist HIGH) + `7BbK` (CodeRabbit Major) — exec fail-open.** A bare failed `exec` exits a non-interactive shell with 127/126 (a non-2 code Claude Code treats as **non-blocking**), so a missing/broken `python3` **silently disabled** the opt-in gate. (`exec … || exit 2` does **not** fix it — a failed exec exits before `||` runs; verified.) Now `shopt -s execfail` makes a failed exec *return*, then an explicit `exit 2` (block sentinel) fires → no `python3` ⇒ writes **blocked**, not allowed. Verified: a PATH without `python3` → hook exits 2.
- **`6_xU` (Codex medium) — unbounded `stdin.read()`.** A huge body could saturate the 5s hook timeout → SIGKILL → non-2 exit → non-blocking. Now `read(4 MiB)`; a larger payload truncates → JSON parse fails → fail closed (ask). +1 test (5 MiB → ask).
- **`7Arr` (Codex P2) + `7BbF` (CodeRabbit Major) — form method-override.** A urlencoded `_method=%44%45%4c%45%54%45` evaded the raw substring scan. Now `parse_qs` (decodes) + an `unquote_plus` regex catch-all. +1 test (URL-encoded `_method=DELETE` → ask).
- **Defense-in-depth:** `main()` wraps the logic so any *unexpected* error fails **closed** (ask), never a non-2 crash. Every path is now exit-0-ask, exit-0-allow, or exit-2-block.

### ⏸️ Declined-with-reason
- **`6_xT` (Codex medium, hardcoded `SCAN_TOOL`).** The tool name is necessarily a static string in the settings **matcher** (the authority — hardcoded there + in `defaultPreToolUseHooks`, and pinned by the contract tests); the impl's check is a defensive mirror. A standalone python hook can't import the JS tool registry, so "registry-driven" doesn't apply.
- **`7Aru` (Codex P2, `auth_profile` override headers).** The hook sees only caller-supplied `tool_input.headers`, **not** the `auth_profile` headers `httpScan` merges *later* (`http-scan.js:196-204`) — it structurally cannot inspect them. The proper fix is runtime-side (strip method-override keys at the profile merge, or reject in `bob_auth_store`) — **tracked as a follow-up**. Low likelihood: profiles store auth headers, not method-override headers.

`test:hooks` write-confirm = **28 cases**; `test:mcp` 2823/0, `test:prompts` 181/0, `test:install` 15/0, `test:cli` 23/0; committed-tree pack 3.225 MB < 3.3 MB.

@coderabbitai review
@codex review

PR #136, opened 2026-06-20
fix(offense): demote IDOR oracle gates #13/#14 to confidence signals (recall 16.7%→50%)

vmihalis:
## What

Demote the two over-strict IDOR-oracle provability gates (#13 `own_scope_not_private`, #14 `identities_collided_not_provable`) from hard refutations to **non-blocking `confidence_signals`**, tripling the oracle's recall (16.7% → 50%) without losing its 0-false-mint precision.

## Why

The first live offensive-measurement run (vu.nl) never fired the offensive producers, and the offline `scripts/measure-idor-oracle.js` harness explained why the IDOR oracle is so quiet: it is **sound but narrow — RECALL 1/6 (16.7%), PRECISION 100%**. The two gates kill a *real* A-only cross-tenant read purely because the response bodies do not echo a private owning-scope key or a distinguishable tenant key.

The canary witness already proves A read B's specific private object while anon (P4) and a third authenticated tenant C (P5) are denied — a categorical cross-principal break. #13/#14 only *strengthen the labeling* of that break as cross-**tenant**; their absence weakens attribution but does not refute the IDOR.

## How

- **`offensive-idor-producer.js`**: gates #13/#14 record a `confidence_signal` and continue to the mint instead of `return fail(...)`. Critically, three values that were hardcoded `true` / assumed tenant presence are now honestly computed:
  - `relationBooleans.tenants_distinct` ← `tenantsProvablyDistinct`
  - `relationBooleans.own_scope_private` ← `ownScopePrivate`
  - `tenant_key: tenantB ? tenantB.key : null` (was `tenantB.key` → **NPE** when #14 trips)

  …so a soft-gated fire never signs a false "distinct/private = true" into the MAC-covered row. `confidence_signals` fold into the `diagnosticBundle` (→ `stderr_hash`) so they are **hash-bound** and survive into the claim with **no MAC change**, and are surfaced on the mint return + `masked_oracle` + the `blocked()` default.
- **Severity unchanged** — `demonstrated_severity` stays `"medium"` (operator decision: Option A). The signals flag the weaker proof; the existing 3-round blind verification + grader corroborate or downgrade.
- **Precision protected by construction** — controls C1/C2 trip *different, upstream* gates (`canary_leaked_unauthenticated` / `p5_authenticated_shared`) that are **not** demoted.

## Scope guard (not touched)

`p5_authenticated_shared` (the C-corroboration reframe is deferred), `canary_leaked_unauthenticated`, `offensive-row-mac.js`, and the core canary-differential mint conditions.

## Verification

```
node scripts/measure-idor-oracle.js
  RECALL 3/6 (50.0%)  PRECISION 100.0%  0 control mints  MODEL ACCURACY 9/9
  CONFIDENCE-SIGNAL CHECK: T6 ✓ T7 ✓   REGRESSION GATE: ✅ PASS (exit 0)
node --test test/offensive-idor-producer.test.js   → 98/98
npm run test:mcp                                    → 2824 pass / 0 fail
```

Tests: inverted the 3 gate negatives (now mint + carry the right signals; the `account_id` case's ledger-absent assertion flips to present), pinned `confidence_signals:[]` on the clean positive, and added a round-trip test proving the signals are physically in the frozen stderr capture and re-hash to the signed `stderr_hash`.

## Summary by CodeRabbit

* **New Features**
  * Added an offline oracle measurement harness with deterministic synthetic fixtures, plus a regression gate to ensure expected PASS/KO behavior stays stable.

* **Bug Fixes**
  * Updated mint-condition logic to replace some hard refutations with confidence-signal-based demotion (including LOW severity when signals are present).
  * Improved owning-scope and same-tenant collision detection, and extended diagnostics to include confidence signals and tenant key context.

* **Tests**
  * Revised AC-6 expectations and added durability/hash-binding and confidence-signal preservation checks.


vmihalis:
@codex review

vmihalis:
@coderabbitai review

vmihalis:
Addressed @codex's P2 ("Block mints when tenants are not distinct") in `0e83ffc`.

The demotion of #14 was too broad — it lumped **provably-same-tenant** (A and B echo the *same* `owner_scope` value) together with the genuinely **unprovable** cases. Provably-same is positive evidence *against* a cross-tenant break (a real same-tenant user-level BOLA would need a different proof), so it must NOT mint as this producer's cross-tenant IDOR.

Fix — split #14:
- **provably same** (both discriminators present at the same key, equal values) → stays a hard `fail("blocked_by_design", "identities_collided_same_tenant")` (no row signed).
- **unprovable** (missing tenant key on either side, or mismatched keys) → records the `identities_collided_not_provable` confidence signal and mints.

Recall is unchanged — the T6/T7 recall-win fixtures are *unprovable*, not same-tenant. Added harness control **C3** (provably-same-tenant must NOT mint) to lock it in the precision measurement:

```
RECALL 3/6 (50.0%)  PRECISION 100.0% (0/3 control mints)  MODEL ACCURACY 10/10
C3 CONTROL: PROVABLY same tenant → identities_collided_same_tenant ✓ declined
```

vmihalis:
@codex review

vmihalis:
@coderabbitai review

vmihalis:
Addressed all four of @codex's round-2 findings in `ee621b8`:

1. **Explicit shared scope (P2 @1035)** — an explicit `owner_scope: "public"`/`"default"` is positive evidence the object is *shared*, so it stays a hard `blocked_by_design` / `own_scope_explicitly_shared`. Only a **missing** scope key demotes (renamed `own_scope_missing`).
2. **Same-tenant aliases (P2 @1049)** — `tenantsProvablySame` now compares tenant **values across alias keys** (`org_id:"acme"` vs `tenant_id:"acme"`), so cross-alias same-tenant hard-blocks instead of soft-minting.
3. **Claim-visibility (P2 @1151)** — you're right that the signal never reaches the claim (`exploit_run` drops extra fields, freeze re-hashes only stdout). Rather than the heavier claim-schema route, a soft-gated fire now stamps **`demonstrated_severity: low`** via a *can-only-lower* override on `buildAndSignOffensiveRow`. Severity **is** claim-visible and enforced, so the unproven-attribution caveat is carried where the grader sees it. Fully-proven fires stay medium.
4. **Harness drift (P3 @554)** — the regression gate now hard-fails on **any** `predicted != actual` (a control declining for the wrong gate no longer passes silently). Added control **C4** (explicit shared); cross-alias same-tenant is covered by a new unit test.

```
RECALL 3/6 (50.0%)  PRECISION 100.0% (0/4 control mints)  MODEL ACCURACY 11/11
C3 same-tenant → identities_collided_same_tenant ✓   C4 explicit-shared → own_scope_explicitly_shared ✓
```
Suites: producer 99/99, test:mcp 2825/0.

vmihalis:
@codex review

vmihalis:
@coderabbitai review

vmihalis:
Addressed the brutalist 🟡 medium (cross-key same-value tenant bypass) in the latest commit — the same-tenant guard now compares the FULL set of owning-scope values across every alias key (case-insensitive), so a shared tenant value under a SECONDARY key (A `owner_scope:"x" + tenant_id:"acme"` vs B `org_id:"acme"`) hard-blocks. New unit test locks it. @codex review

vmihalis:
Round 4: addressed the P2 "Honor shared scope from owner readback". #13 now hard-blocks `own_scope_explicitly_shared` when B's trusted owner readback carries an explicit shared scope, even if the live P1 GET omits owning-scope keys. Asymmetric by the absence-vs-positive rule (the readback can only strengthen the block, never clear the P1 soft-gate). +2 producer tests, +C5 harness control. RECALL 3/6 held, PRECISION 100% (0/5 controls), test:mcp green.

@codex review
@coderabbitai review

vmihalis:
Round 5: addressed three of the four P2s — all the same theme (consider every owning-scope alias across all proof bodies, not just P1's first alias):

- **Check every readback scope alias** — `#13` now scans every `OWNING_SCOPE_KEYS` alias (new `hasExplicitSharedScope` helper), so `{owner_scope:"tenant-B", workspace_id:"public"}` no longer hides a shared label behind a private primary alias.
- **Block shared scope echoed in proof reads** — the shared-scope scan now runs over `[p1Parsed, parsedReadback, p2Parsed, p2primeParsed]`, so an explicit `owner_scope:"public"` in the *signed A→B proof body* hard-blocks even when the readback is private and P1 omits the key.
- **Fold owner readback into same-tenant guard** — `#14` `bScopeValues` is now the union of P1 + readback owning-scope values, so a readback B-scope that collides with A's own scope (P3) hard-blocks `identities_collided_same_tenant`. (Asymmetric as before — folding the readback only *adds* a refutation; differing values do not clear the P1-side distinctness soft-gate.)

**On "Require stderr capture for confidence signals" — declining for this PR, with justification:** the security-relevant consequence of the `#13/#14` signals is the **LOW severity ceiling**, and that is already claim-visible and MAC-bound — `demonstrated_severity:"low"` is stamped into the signed offensive row and bound through the claim (`offensiveRunRowSatisfiesEvidence`), and the row MAC covers `stderr_hash` itself. So deleting/tampering `<run_id>.stderr` **cannot upgrade severity or forge a mint**; it only removes verifier-facing signal *detail* whose hash is still MAC-bound in the row. Adding stderr-file re-validation to the audit-graded `claim-freeze.js` completeness path is a reasonable follow-up but is a separate, broader subsystem change beyond this producer PR and is not required for the soundness of the severity decision. Happy to file it as a tracked follow-up.

Tests +3 (secondary-alias readback; P2/P2′ proof read shared; readback-vs-A same-tenant). Harness +C6/+C7. RECALL 3/6 held, PRECISION 100% (0/7 controls), MODEL 14/14, test:mcp 2831/0.

@codex review
@coderabbitai review

vmihalis:
Round 6:

- **Integer tenant IDs (MEDIUM)** — `ownScopeOf`/`tenantDiscriminator`/`owningScopeValues` required `typeof === "string"`, so a numeric `org_id`/`tenant_id` (common in real REST APIs) was invisible and a same-tenant BOLA (A & B both `org_id:7`) could be minted as cross-tenant. New `scopeValueString()` coerces a finite number to its string form; all three readers use it now, so numeric IDs feed the same-tenant + shared-scope guards.
- **Harness not in CI (process headline)** — wired `measure-idor-oracle.js` into the suite: `test/measure-idor-oracle-gate.test.js` spawns it and asserts exit 0 + "REGRESSION GATE PASS", registered in the mcp test manifest. A control false-mint or a predicted-signal break now fails CI, so the empirical foundation can't rot silently.
- **Severity clamp -1 (LOW)** — the `Math.max(overrideIdx, indexOf(registrySeverity))` clamp now validates the registry index and fails closed (a registry typo would otherwise silently elect the override regardless of direction).
- **`blocked()` nit** — `confidence_signals:[]` is spread after `...extra`, so a hard block can never carry a soft-gate signal.

**Declining (Codex architectural):** MAC-signed soft-gated rows legally carrying `tenants_distinct:false` is the intended severity-as-signal design (operator-approved) — the relation booleans honestly reflect the unproven attribution and the LOW severity (claim-visible, MAC-bound) carries the consequence into grade.

+1 producer test (integer same-tenant), +C8 control, +1 harness-gate test. RECALL 3/6 held, PRECISION 100% (0/8 controls), MODEL 15/15, test:mcp 2833/0.

@codex review
@coderabbitai review

vmihalis:
Round 7: `scopeValueString` now uses `Number.isSafeInteger` instead of `Number.isFinite`, so a 64-bit numeric tenant id above 2^53 (already precision-lost by `JSON.parse`) is no longer stringified to an unreliable value that could collide two distinct tenants — it falls to `null` (no usable scope value) and the read soft-gates rather than risk a precision-wrong same-tenant verdict. Safe integers still coerce. Extended the `tenantDiscriminator` unit test. RECALL 3/6 held, PRECISION 100% (0/8), test:mcp 2833/0.

@codex review
@coderabbitai review

vmihalis:
Round 8: added DIRECT unit tests for the `demonstratedSeverityOverride` clamp in `offensive-capture-writer.test.js` — a fails-closed guard (unknown override throws before any write) and a full-sign behavioral test (override `low` lowers medium→low, `critical` does NOT raise above medium, no override keeps the ceiling). Locks the "can only lower, never raise" contract the producer itself never exercises. capture-writer 7/7, test:mcp 2835/0.

On the two design observations — signed soft-gated rows carrying `tenants_distinct:false`/`own_scope_private:false`, and soft-gated fires capping permanently at LOW with no auto-upgrade — these are **declined as the intended, operator-approved Option-B1 design**: the confidence signal is not in the claim/grade schema, so severity is the claim-visible carrier of the unproven attribution (operator chose LOW for weak fires). The signed row stays honest about what the producer actually proved; a human can file a confirmed cross-tenant break separately.

@codex review
@coderabbitai review

vmihalis:
Round 9 — both P1 false-mint paths + the medium THROW-gate weakness (2f54c07 brutalist) fixed, plus the cheap sound folds:

- **Unsafe-magnitude numeric scope soft-mint (Codex P1)** — round 7's `Number.isSafeInteger` guard drops a *present-but-unsafe* numeric owning-scope (a 64-bit `org_id` above 2^53, already precision-lost by `JSON.parse`) to `null`, which the same-tenant guard read as "scope ABSENT" → two SAME-tenant objects could soft-mint as cross-tenant. New `hasUnusableOwningScope()` **hard-blocks `own_scope_unusable_numeric`** across every body feeding the tenant guard (B side p1/readback/p2/p2′ AND A side p3): a present-but-unsafe discriminator can't prove *distinct* tenants, so fail closed instead of downgrading to "missing".
- **Proof-read same-tenant not folded (Codex P1)** — `bScopeValues` folded only P1 + readback, so when both omit B's scope but the signed A→B proof bodies (P2/P2′) themselves reveal O_B sits in A's tenant, the collision was missed → soft-mint. P2/P2′ are already trusted for the explicit-shared hard block, so they now **also feed the positive same-tenant refutation**. This can only ADD a refutation (a real cross-tenant proof carries B's tenant ≠ A), never a false collision.
- **`matches()` accepted any THROW (Codex/Claude/brutalist medium)** — a prefix match let T4 pass on *any* thrown error, so a regressed guard throwing a different error first would silently stop exercising the invariant T4 names. Now `matches()` is **exact on the THROW code**, T4 pins `THROW:INVALID_ARGUMENTS`, and a new `predicted_error_includes` field pins the *specific* reason (`"must not include a query string"`) via a THROW-REASON gate check.

Cheap sound folds (the three sub-threshold lows):
- Registry-severity membership now validated on **every path** in `offensive-capture-writer` (not only when an override is supplied) — a registry typo can't pass unvalidated to a signed row on the no-override path.
- The two `confidence_signals` return slots (top-level + `masked_oracle`) each get an independent `.slice()`.
- **Documented the Option-B1 boundary** at `offensiveRunRowSatisfiesEvidence` (claims.js): signals are advisory forensics carried by severity, intentionally not a schema-enforced gate.

**Declining (re-stated, operator-approved):** the claim/grade schema not structurally enforcing `confidence_signals` is the intended Option-B1 design — severity is the claim-visible carrier of weak attribution; the signed row stays honest about what was proven (now documented at the exact validator Codex flagged).

Tests: +2 producer unit tests + 2 harness controls (C9/C10) + the THROW-REASON gate check. **RECALL 3/6 held, PRECISION 100% (0 false mints on 10 controls), MODEL ACCURACY 17/17**, producer 108/108, test:mcp 2837/0.

@codex review
@coderabbitai review

vmihalis:
fix(offense): close the 3 round-9 brutalist mediums — clamp invariant, case-symmetry, harness wiring (PR#136 round 10)

1. Math.max clamp invariant (Claude) — `SEVERITY_VALUES[Math.max(overrideIdx, registryIdx)]` only
   LOWERS severity because SEVERITY_VALUES is descending (most-severe first); no test enforced that, so
   a reorder would silently invert the clamp (an override could RAISE the signed ceiling). Added an
   explicit test asserting SEVERITY_VALUES stays `["critical","high","medium","low","info"]` — a
   reorder now fails loudly here, not by mis-signing a row in production.

2. Case-sensitivity asymmetry (Claude) — the same-tenant HARD block compares lowercased values
   (owningScopeValues) but tenantsProvablyDistinct compared raw case-sensitively, so a case-only
   difference ("Acme" vs "acme") could read as "provably distinct" while the same-tenant guard already
   treats it as one tenant — a false-negative dead-zone. tenantsProvablyDistinct now compares
   case-insensitively too (the safe direction: prefer same-tenant -> no mint). Consistent on both legs.

3. Harness discoverability (Codex) — added `npm run measure:idor`. The regression gate already runs in
   CI (test/measure-idor-oracle-gate.test.js is in the mcp manifest, so test:mcp spawns the harness and
   fails on a control false-mint or signal/throw-reason mismatch); the script is now also directly
   runnable for humans.

measure:idor RECALL 3/6, PRECISION 100% (0/10 controls), MODEL 17/17; capture-writer 8/8, producer
108/108, test:mcp 2838/0.

@codex review
@coderabbitai review

vmihalis:
fix(offense): nested-envelope scope scanning — close the soft-mint amplification (PR#136 round 11)

Round-10 brutalist convergent finding (Codex correctness + Claude correctness): the scope-detection
readers (ownScopeOf, tenantDiscriminator, owningScopeValues, hasUnusableOwningScope, and via them
hasExplicitSharedScope) only read TOP-LEVEL body keys. Many real REST APIs nest the resource one level
under a {data:{...}} / {result:{...}} / JSON:API {data:{attributes:{...}}} envelope, so the owning
scope is invisible. Before this PR that was a hard block (safe); after the #13/#14 demotion, a
nested-envelope body with no top-level scope soft-gates -> a SAME-tenant nested API could soft-mint a
LOW cross-tenant row. This PR introduced that amplification, so it fixes it here.

Fix: a bounded scopeSearchRoots() scans the TOP-LEVEL body AND one level into common envelope keys
(data, result, item, record, attributes, payload, resource); all four scope readers iterate it. The
top-level root is ALWAYS scanned first (flat bodies unchanged), arrays are not descended, and only ONE
level deep is scanned. The direction is precision-safe: more scope detected = more same-tenant /
explicit-shared / unusable-numeric HARD blocks, never a new mint.

Verified: corpus UNCHANGED (RECALL 3/6, PRECISION 100%, MODEL 17/17 -> 18/18 with the new control);
a nested {data:{org_id:7}} same-tenant pair now hard-blocks identities_collided_same_tenant instead of
soft-minting. +1 producer test, +1 harness control (C11). producer 109/109, test:mcp 2839/0.

Declining (re-stated, operator-approved Option-B1, logged across rounds 6-10): the claim gate not
structurally enforcing confidence_signals / the signed relation booleans (tenants_distinct:false,
own_scope_private:false) / demonstratedSeverityOverride lacking a machine-checkable reason field — these
all encode the same intentional design: a soft-gated fire is claim-visible via its LOWERED severity
(the only field the claim/grade path carries), the relation booleans honestly reflect what was proven,
and the confidence_signals are hash-bound advisory forensics (documented at offensiveRunRowSatisfiesEvidence).
The canary witness still proves a real cross-PRINCIPAL access; LOW + signals flag the unproven
cross-TENANT attribution. A reason-bound severity-downgrade schema is the deferred Option-B2 seam.

@codex review
@coderabbitai review

vmihalis:
Round 12 — Codex P1 fixed: scopeSearchRoots now descends TWO levels via the well-known wrapper keys so JSON:API (data -> attributes -> org_id) is reached; bounded to 2 levels, dedup-guarded, precision-safe. Corpus unchanged (RECALL 3/6, PRECISION 100

vmihalis:
Round 12 — Codex P1 fixed: `scopeSearchRoots` now descends TWO levels via the well-known wrapper keys so JSON:API (`data` → `attributes` → `org_id`) is reached; bounded to exactly 2 levels, dedup-guarded, precision-safe (more scope detected = more HARD blocks, never a new mint). Verified flat / one-level / JSON:API two-level all resolve the scope, and a 3-level body resolves to null (bounded out).

Corpus unchanged (RECALL 3/6, PRECISION 100 percent, MODEL 18/18 → 19/19); a JSON:API same-tenant pair now hard-blocks `identities_collided_same_tenant`. +1 producer test, +1 harness control (C12). producer 110/110, test:mcp 2840/0.

@codex review
@coderabbitai review

vmihalis:
Round-12 review — the 3 remaining threads are all the (operator-approved, logged) Option-B1 design point plus one naming nit; declining with justification and converging.

**Declining (Codex design, claims.js:880 + capture-writer.js:200):** "attribution confidence lives only in `demonstrated_severity`, not a schema-enforced field" / "`demonstratedSeverityOverride` turns a registry-authoritative decision into a producer convention" — this is the intended Option-B1 design, declined and logged every round since round 6 and documented in-code at `offensiveRunRowSatisfiesEvidence` (claims.js): a soft-gated fire is claim-visible via its LOWERED severity (the only field the claim/grade path carries), the signed relation booleans honestly reflect what was proven, and `confidence_signals[]` are hash-bound advisory forensics. The override is registry-BOUNDED — it can only ever LOWER `OFFENSIVE_TOOL_DEMONSTRATED_CEILING` (Math.max over descending SEVERITY_VALUES, fail-closed, with the descending-order invariant now locked by a test), so the registry stays the ceiling authority; the producer can only demote within it. A reason-bound severity-downgrade schema is the deferred Option-B2 seam.

**Declining (Claude maintainability, capture-writer.js:249):** the `demonstratedSeverityOverride` name vs the can-only-lower clamp — the behavior is documented immediately above the clamp ("An optional per-row override may ONLY LOWER the registry ceiling ... It can never RAISE severity") and locked by the round-10 SEVERITY_VALUES-descending test and the round-8 behavioral tests (override "critical" cannot raise above the medium ceiling). A rename would be churn across the producer + writer for a 12-round PR with the semantics already pinned by tests + comment; deferring.

Net for the PR: IDOR oracle recall 16.7%→50%, precision 100% (0 false mints on 12 controls incl. unsafe-numeric / proof-read / nested-envelope / JSON:API same-tenant), MODEL 19/19; every concrete finding across 12 rounds fixed. test:mcp 2840/0.

PR #138, opened 2026-06-20
fix(session): derive auth_status on advance so it reflects reality (Option C)

vmihalis:
## What

Make `auth_status` meaningful (operator's **Option C**): derive it on every lifecycle advance instead of leaving it stuck at `"pending"` forever.

## Why

`auth_status` was initialized to `"pending"` and had **no setter anywhere** — `bob_advance_session` carried the prior `auth_context` forward unchanged and **ignored** the `auth_status` arg the orchestrator skill passes. So a session that authenticated (a pasted cookie / stored profile) still reported `"pending"`, misleading the grader/operator about whether authenticated testing happened (exactly what we saw on the vu.nl run — the attacker profile got 200s but `auth_status` stayed `pending`).

## How

`advanceSession` now derives `auth_status` at the nucleus rebuild. Precedence:

1. **Explicit** caller-supplied `auth_status` wins — now honored (this also makes `--no-auth`'s `"unauthenticated"` actually take effect, where it was previously dropped).
2. Else, if a **usable auth profile is stored** (`auth.js` `hasUsableAuthProfile`) → `"authenticated"`.
3. Else **carry the prior status forward** — only the explicit path ever sets `"unauthenticated"`; never auto-downgrade.

`buildSessionNucleus → normalizeAuthContext` validates the resulting enum, and the read surfaces (session/state summary, pipeline-analytics) read the nucleus, so they reflect it automatically — no reader changes needed.

## Files

- `auth.js`: `hasUsableAuthProfile(domain)` — true iff `auth.json` carries ≥1 stored profile (reuses the same candidate-domain + migrate read as `listAuthProfiles`; no new coupling, no circular import).
- `session-state.js`: `deriveAdvanceAuthContext` wired into `advanceSession`.
- `test/lifecycle-advance.test.js`: 3 precedence tests (derive authenticated, carry pending, explicit wins).

## Verification

```
npm run test:mcp → 2826 pass / 0 fail
```

## Summary by CodeRabbit

* **New Features**
  * Session lifecycle advancement now recalculates `auth_context.auth_status` every step and keeps it in sync with `state.json`.
  * Adds an audit/governance event (`auth_context.replaced`) whenever the normalized `auth_status` value changes.

* **Bug Fixes**
  * Refines `auth_status` derivation and precedence: usable stored credentials can promote to authenticated, operator-forced overrides take priority, `"unauthenticated"` remains sticky, and explicit `"authenticated"` is ignored without usable stored credentials.
  * Rejects invalid non-blank `auth_status`; normalizes whitespace/empty inputs as omitted.

* **Tests**
  * Expanded coverage for precedence, forge-guard behavior, normalization, and boundary validation.

* **Documentation**
  * Updates session-advance tool contract: canonical accepts optional `auth_status` override; legacy alias accepts it but does not forward it.


vmihalis:
@codex review

vmihalis:
@coderabbitai review

vmihalis:
Addressed all three findings (Codex P2 ×3 + brutalist) in the latest commit:
1. **Explicit path dead via dispatch** — added `auth_status` (AUTH_STATUS_VALUES enum) to the canonical `bob_advance_session` inputSchema (and forwarded it through the legacy alias), so the explicit path (and `--no-auth`'s `unauthenticated`) is now reachable.
2. **state.json split-brain** — `nextState` now mirrors `nextAuthContext.auth_status` so state.json and the nucleus stay in lockstep.
3. **Empty profile = usable** — `hasUsableAuthProfile` now requires at least one real credential (a non-metadata key); an empty/metadata-only profile no longer marks the session authenticated.
Tests added (state.json mirror + empty-profile-stays-pending). test:mcp 2827/0. @codex review

vmihalis:
@coderabbitai review

vmihalis:
Round 2: addressed the converged Codex + CodeRabbit + brutalist findings on the latest commit.

- **`hasUsableAuthProfile` exclusion-list → positive-inclusion** (Codex + Claude MEDIUM): now an allowlist — a usable profile must carry an outbound session-credential header (`Authorization`/`Cookie`) with a non-empty value, so a future profile field can't silently promote a session to `authenticated`. A bespoke-header-only profile reads as not-yet-authenticated (the safe false-negative direction).
- **Explicit-override forge guard** (Codex MEDIUM/security): an explicit `auth_status: "authenticated"` is now honored only when a usable profile is present OR the caller carries `operator_force`. Negative/neutral assertions (`unauthenticated` via `--no-auth`, `pending`) still need no evidence.
- **Sticky `--no-auth`** (Codex P2): a prior explicit `unauthenticated` is carried forward across a later advance that omits `auth_status`, even when a captured profile exists.
- **Milestone semantics documented** (Claude MEDIUM/design): `auth_status` is a session milestone (recomputed at advance time, sticky, mid-advance credential deletion not auto-detected), not a real-time probe; consumers needing live presence call `hasUsableAuthProfile`.
- **Legacy adapter** (CodeRabbit Minor + Codex LOW): trim + only forward non-empty `auth_status`; legacy enum aligned to `AUTH_STATUS_VALUES`.

On the **FS-I/O-on-every-advance** LOW (Codex): accepted as-is — `hasUsableAuthProfile` is one cached `auth.json` read per *lifecycle advance* (a handful per session, not per request), and reading profile presence at advance time is exactly what makes the derived status reflect reality. Not worth coupling-avoidance churn.

Tests +4 (non-cred-header stays pending; `--no-auth` sticky across a later advance; unbacked explicit `authenticated` ignored; `operator_force` `authenticated` honored). lifecycle-advance 22/22, test:mcp 1071/0, test:prompts 181/0.

@codex review
@coderabbitai review

vmihalis:
Round 3:

- **Empty-string split-brain (headline)** — fixed both ways: `deriveAdvanceAuthContext` now treats a blank/whitespace `auth_status` as omitted (can't win precedence (1)), AND the `state.json` mirror now copies the **normalized** `nextNucleus.auth_context.auth_status` rather than the pre-normalization derive output, so the nucleus and state.json agree by construction for any value the nucleus rewrites. +1 test (blank arg → omitted, nucleus == state.json).
- **Forge-guard comment** — reworded: the guard stops an *unprivileged* caller (no `operator_force`) like a prompt-injected evaluator; `operator_force` is the deliberate operator escape hatch (an operator vouching for an out-of-band context CAN assert `authenticated`).
- **Nit** — dropped the redundant `String(key)` cast (`Object.entries` keys are strings).

**Declining, with justification:**
- *session→auth coupling* (Codex design): `session-state` importing `hasUsableAuthProfile` is appropriate layering — the session lifecycle legitimately derives status from auth state; pushing a resolved boolean up to the tool handler just relocates the same dependency.
- *FS read in the session lock* (Codex perf LOW): one cached `auth.json` read per *lifecycle advance* (a handful per session), and reading presence at advance time is exactly what makes the status reflect reality.
- *rule (4) `|| "pending"` fallback* (Claude LOW): a safe corrupt→pending default, and moot now that mirrored values are normalized.

lifecycle-advance 23/23, test:mcp 2832/0.

@codex review
@coderabbitai review

vmihalis:
Round 4:

- **Sticky `--no-auth` headline (Claude MEDIUM)** — fixed. The guarantee was only holding against an *omitted* `auth_status`; an unprivileged explicit `authenticated` with a matching profile in the store (e.g. a victim credential captured for IDOR testing) silently reversed a prior `--no-auth`. Restructured the precedence so it's coherent: **(1)** `operator_force` = authority (explicit always honored, incl. lifting a prior unauthenticated); **(2)** a prior `unauthenticated` is sticky and now survives *both* omission *and* an unprivileged backed `authenticated` — only operator authority lifts it; **(3)** unprivileged explicit (negative/neutral always, positive only when profile-backed); **(4)/(5)** derive / carry forward. Exported `deriveAdvanceAuthContext` + a direct 10-case precedence-table unit test.

**Declining, with justification:**
- *bare `auth.json` catch* (Claude): fail-closed by design — an unreadable profile means "no usable credential" (the safe default); surfacing the IO error here would couple the lifecycle to auth-store health, which the same review flagged as bad coupling elsewhere.
- *profile-name blindness in multi-persona* (Codex): `auth_status` is a session-level signal by design, and the multi-persona hazard (a victim credential promoting the session) is now mitigated by the sticky `--no-auth` fix above.
- *explicit `pending` reset* (Claude): a benign neutral re-assertion.

lifecycle-advance 24/24, test:mcp 2833/0.

@codex review
@coderabbitai review

vmihalis:
Round 5 — three MEDIUMs addressed:

- **Governance event ordering (Claude headline)** — moved the `deriveAdvanceAuthContext` call (its `hasUsableAuthProfile` auth.json read is the only remaining fallible work) AHEAD of the `operator_force` override audit append, so a read failure can no longer orphan a phantom override event ("all fallible work before durable writes").
- **Legacy-adapter authority-conflation (Codex security)** — dropped the `auth_status` forwarding from the `bounty_transition_phase` adapter entirely (restoring its original "ignored" contract). Since the shim maps `override_reason`→`operator_force` (auth authority), forwarding `auth_status` let a lifecycle-bypass caller manufacture operator-level auth provenance. Auth context now flows only through the canonical tool. +test (adapter drops `auth_status`).
- **Boundary validation (Claude)** — `auth_status` is now validated at the call boundary (clear `INVALID_ARGUMENTS`), not deep in `normalizeAuthContext`; blank still treated as omitted. +test.

**Declining (restated, justified):** sticky-authenticated after credential deletion (deliberate milestone design, documented); synchronous auth I/O per advance (bounded — one cached read per lifecycle transition); bespoke API-key headers excluded from positive-inclusion (safe false-negative — session stays pending, auth tasks run); `deriveAdvanceAuthContext` now public (intentional, backs the precedence unit test).

lifecycle-advance 26/26, test:mcp 2833/0.

@codex review
@coderabbitai review

vmihalis:
Round 6:

- **Padded `auth_status` (Codex P2)** — fixed. The round-5 boundary check trimmed for validation but `deriveAdvanceAuthContext` used the RAW value, so a padded `"  authenticated  "` passed the boundary yet threw deep in `normalizeAuthContext` on the padded enum value. Now the non-blank explicit value is trimmed before derivation, so it's honored as its trimmed form and reaches the nucleus enum check trimmed. +1 test.

**Declining (justified):** "honor profile expiry before promoting `auth_status`" (Codex P2) — `auth_status` is a documented session **milestone** (highest auth state reached, sticky, not a real-time validity probe), so a credential usable when an advance ran legitimately set the milestone. Coupling `hasUsableAuthProfile` to credential-freshness (parsing the optional, format-varied `expires_at`/`expiry` hints + a time comparison) would contradict the field's stated semantics; a consumer needing live validity checks the credential directly (a stale cookie 401s on use regardless of `auth_status`). Noted as a reasonable future enhancement.

lifecycle-advance 27/27, test:mcp 2836/0.

@codex review
@coderabbitai review

vmihalis:
Round 7 — all 3 convergent brutalist mediums (df584cf) addressed:

- **Two-step forge path (Codex, headline)** — the forge guard only blocks a *bare* unprivileged explicit `"authenticated"`; an evaluator that calls the pre-approved `bob_auth_store` can stash a credential-shaped header and reach the profile-backed derivation, so the comment's "cannot fabricate credential provenance" was an overclaim. Fixed two ways: (a) **scoped the comment honestly** — `auth_status` is an advisory *milestone* that grants no capability (a forged token 401s on use), and (b) **now emit a `governance.auth_context.replaced` audit event on every `auth_status` change**, so a fabricated milestone is always reconstructable from the lifecycle log. This also closes the second convergent point (no dedicated governance event for auth changes).
- **Exported `deriveAdvanceAuthContext` lacked internal validation (Claude)** — it is module-exported and callable in-process without `advanceSession`'s boundary check, so an invalid explicit value would propagate to a deep `normalizeAuthContext` throw. Added a **self-guard**: a non-blank value outside `AUTH_STATUS_VALUES` throws a clear `INVALID_ARGUMENTS` at the function boundary; blank stays allowed (omitted).
- **Unprivileged `"pending"` downgrade (Claude)** — branch (3) honored any non-`"authenticated"` explicit, so explicit `"pending"` could regress an established `"authenticated"` back to "unknown". Added a **no-regress guard**: an explicit `"pending"` never downgrades a non-pending prior (falls through to carry-forward). The forge guard already blocked the elevation direction; this blocks the regress.

**Declining (justified):** Codex "schema inconsistency between `bob_auth_store`'s `credentials` field and `CREDENTIAL_HEADER_NAMES`" — `bob_auth_store`'s `credentials` is `{email, password}` (**login material**), semantically distinct from a session credential header (`authorization`/`cookie`). A profile carrying only login material has not yet exchanged it for a session token, so `hasUsableAuthProfile` correctly returns `false` and `auth_status` stays `"pending"` until a token header is stored. This is the documented safe-direction (`auth.js:407-409`), not an inconsistency.

+2 tests (self-guard throws on invalid exported call; `auth_context.replaced` fires on change only); extended the precedence table with the no-regress assertions. lifecycle-advance 29/29, test:mcp 2838/0.

@codex review
@coderabbitai review

vmihalis:
fix(session): non-string self-guard + atomic auth-status provenance (PR#138 round 8)

Round-7 review (1c00d29) raised 2 legitimate fixes and 2 design points (declined with justification):

1. Non-string self-guard (CodeRabbit) — the exported deriveAdvanceAuthContext validated the enum but
   silently treated a NON-string explicit value as omitted, diverging from advanceSession's boundary
   (which rejects non-strings). It now rejects a non-null non-string outright with the same clear
   "auth_status must be a string" error. +test (42 → throws; null/omitted still derives).

2. Atomic auth-status provenance (CodeRabbit Major + Codex) — the new governance.auth_context.replaced
   append was a SECOND durable write after the advance had already committed (state.json + nucleus +
   lifecycle.advanced), so a throw there would fail the tool despite a committed state change, and the
   provenance was not atomic with the advance. Fixed two ways: (a) the auth transition (from/to_auth_status
   + auth_status_changed) now rides on the CANONICAL governance.lifecycle.advanced payload, so it is
   captured atomically with the lifecycle move and reconstructable from that single append; (b) the
   dedicated governance.auth_context.replaced companion (kept as a queryable index) is now BEST-EFFORT
   (try/caught, like the pipeline-events mirror), so its failure can never fail an already-committed
   advance. +test asserts the advanced event carries the transition.

Declined (justified, with sharpened docs):
- "hasUsableAuthProfile promotes auth_status for victim/IDOR profiles" (Codex) — name-agnostic
  promotion is intentional (the operator plan advances on an attacker OR victim profile) and
  auth_status is an advisory MILESTONE meaning "the session holds >=1 usable credential profile (any
  principal)", not a per-principal authentication claim. It grants no capability; coupling the
  milestone to free-form caller profile names would be fragile for no security gain. Doc sharpened.
- "operator_force conflates gate-bypass with auth-context authority" (Claude) — deliberate: operator_force
  IS the operator-authority signal and intentionally spans both, defensible for an advisory field whose
  every forced advance already carries an audited override_reason; a separate flag is API surface for no
  gain. An unprivileged caller still cannot forge "authenticated". Doc sharpened.

lifecycle-advance 29/29, test:mcp 2838/0.

@codex review
@coderabbitai review

vmihalis:
Round-8 review triage — both remaining findings declined with justification (logged per the pre-merge gate); converging.

**Declining (Codex P1 "make auth provenance durable before committing state", session-state.js:852):** the concern is that the canonical `governance.lifecycle.advanced` append (which now also carries the auth transition) is post-commit, so a rare append failure (disk full / events file is a directory) throws after the state+nucleus are already written. This is **pre-existing advanceSession behavior unchanged by this PR** — that event has always appended after the durable lifecycle write. Crucially, the auth_status **value** is durably committed in the nucleus *before* the event (a consumer reads the current value from the nucleus, never by replaying events), so this is an audit-event-durability concern, not a state-integrity one. The dedicated `governance.auth_context.replaced` companion is already best-effort; making the *canonical* advanced event best-effort would instead **hide** a real partial-write the operator should see. Reworking advance-event durability to pre-commit is a separate lifecycle refactor out of scope for the auth_status PR.

**Declining (Codex P2 "require a reason for forced auth assertions", session-state.js:556):** coupling `operator_force`'s auth-context authority to a mandatory `override_reason` adds API friction for an **advisory** field that grants no capability (a forged/stale token 401s on use), and every `operator_force` advance is already recorded as a `governance.lifecycle.override` event. A reasonable future hardening, not a correctness gate.

No code change from these declines. lifecycle-advance 29/29, test:mcp 2838/0 (the round-8 fixes: non-string self-guard + atomic auth provenance on the advanced event + best-effort companion).

PR #139, opened 2026-06-20
fix(recon): identity-aware ProjectDiscovery tool resolution (defeat httpx shadowing)

vmihalis:
fix(recon): identity-aware ProjectDiscovery tool resolution (defeat httpx shadowing)

A live run collapsed recon to ~5 surfaces (vs thousands) because /usr/local/bin/httpx — the PYTHON
httpx HTTP-client CLI — shadows the ProjectDiscovery ~/go/bin/httpx host-prober on PATH. The recon
chain resolved httpx with `command -v httpx 2>/dev/null || ...; [ -z "$HTTPX" ] && [ -x ~/go/bin/httpx ]`,
but `command -v` SUCCEEDS with the python shadow, so (a) the ~/go/bin fallback never fired and (b) the
wrong binary ran `httpx -l ... -silent ...`, which the python client doesn't understand → empty
live_hosts → katana + nuclei got empty inputs → the whole pipeline starved silently. The binary-check
likewise reported OK:httpx for the shadow (false positive).

Root-cause fix — resolve every ProjectDiscovery Go tool (httpx, katana, dnsx, tlsx, subzy) to its REAL
compiled binary, preferring ~/go/bin and skipping any same-named #!-script shadow on PATH:

  TOOL=""; for c in "$HOME/go/bin/tool" "$(command -v tool 2>/dev/null||true)"; do \
    [ -n "$c" ] && [ -x "$c" ] && [ "$(head -c2 "$c" 2>/dev/null)" != '#!' ] && { TOOL="$c"; break; }; done

This is host-agnostic: if the PD binary is in ~/go/bin it wins regardless of PATH order; if PATH's
httpx is the python #!-script it is rejected; a real PD binary anywhere (no shadow) still resolves.
jwt_tool keeps its own resolution (it IS legitimately a python script). Verified against all three
scenarios (go/bin+shadow → go/bin; shadow-only → MISSING; PATH-only real → PATH).

Also:
- The binary-check (surface-discovery-tools.txt) now reports MISSING:httpx / MISSING:katana when only a
  #!-script shadow exists, so the operator sees the real install gap instead of a false OK.
- A STARVED recon is now VISIBLE: when real subdomains were enumerated but httpx probed 0 live hosts
  (or httpx was unresolved), a DEGRADED:live_hosts_empty marker is appended before the apex fallback
  masks it — so a collapsed-breadth run no longer reads as a clean "well-hardened, few surfaces" result.

Both surface-discovery and deep-surface-discovery source prompts are fixed; the Claude agents and the
Codex/Kimi adapters are regenerated. test:prompts 181/0.

## Release Notes

* **Bug Fixes**
  * Improved tool discovery reliability by preventing script files from masking compiled binaries on execution paths.
  * Added reporting of incomplete reconnaissance results when host enumeration yields no live targets from discovered subdomains, providing better visibility into coverage gaps.


vmihalis:
Root-cause fix for the recon-starvation regression seen on the live vu.nl run (5 surfaces vs thousands): the ProjectDiscovery `httpx` host-prober was shadowed on PATH by the python `httpx` HTTP-client CLI, and the existing `command -v httpx || ~/go/bin` fallback never fired because `command -v` *succeeds* with the wrong binary. Now every PD Go tool (httpx/katana/dnsx/tlsx/subzy) is resolved identity-aware: prefer `~/go/bin`, skip any same-named `#!`-script shadow. Plus an honest binary-check (MISSING when only a shadow exists) and a `DEGRADED:live_hosts_empty` marker so a starved recon is visible instead of masked by the apex fallback. Resolver verified against go/bin+shadow / shadow-only / PATH-only-real. test:prompts 181/0.

@codex review
@coderabbitai review

vmihalis:
fix(recon): target ONLY the python shadow, not all #!-scripts (PR#139 round 2 — shim-manager safe)

Round-1 brutalist (Codex, headline): the blanket `#!`-script rejection correctly defeats the python
httpx HTTP-client shadow but ALSO rejected every shim-manager install — asdf / mise / nix wrap the real
ProjectDiscovery binary in a SHELL `#!`-script, so those environments would falsely report the tool
MISSING and degrade recon (a regression for a common macOS/dev setup).

Fix: narrow the discriminator from "is a #!-script" to "is a python-interpreter #!-script". A candidate
is skipped ONLY when its shebang names python (the httpx HTTP client); compiled binaries (direct
go install, Homebrew symlink) AND shell shims (asdf/mise/nix proxying to the real PD binary) are
accepted:

  ! { [ "$(head -c2 "$c")" = '#!' ] && head -1 "$c" | grep -qi python; }

Verified: python-shebang script -> reject; bash/sh shim -> ACCEPT; compiled ELF -> ACCEPT. Applied to
all httpx/katana/dnsx/tlsx/subzy resolvers and both binary-checks in surface-discovery and
deep-surface-discovery; Claude agents + Codex/Kimi adapters regenerated. test:prompts 181/0.

Declined (justified): the family-probe path emitting its own DEGRADED marker (Claude medium) — the
main live-host step already records `httpx_unresolved` when no PD binary resolves, and the family
probe resolves the SAME binary on the SAME host, so a second marker would be redundant noise for the
one failure mode that matters (binary unresolved). The main-path marker covers both probes.

@codex review
@coderabbitai review

PR #144, opened 2026-06-20
feat(evidence): carry a real proof-of-impact sample in authorized engagements

vmihalis:
## What

Lets the evidence pack carry a **real proof-of-impact sample** for data-exposure findings in **authorized engagements**, instead of always masking it down to redacted IDs + field names.

## Why

Audit question: *what stops Bob from showing actual leaked data as proof?* Result — almost nothing in code does; the suppression was one prompt. `prompts/roles/evidence.md` (→ `evidence-agent.md`) hard-mandated "redact … full PII values, raw … response bodies" and `redacted_object_id`. So even when an evaluator pulls a real cross-scope sample, the evidence pack feeding the report stripped it. For an authorized client/asset-owner engagement that **undersells impact** — the data owner is paying to see the real records; a masked count won't land.

## Change

Split the redaction rule by what is *actually* sensitive:
- **Always redact** secrets, auth headers, cookies, tokens, passwords, API keys, SC bytecode — these never belong in a report, any engagement. *(unchanged)*
- **Data-exposure proof-of-impact:** in an authorized engagement (client/asset-owner, or `BOB_DATA_ACCESS_AUTHORIZED`) include a **small real sample (≤10 records)** with the real exposed values; on a **public safe-harbor** program that caps data access, redact PII + pair with schema + count. **Never the full corpus** — a bounded sample, not a dump.
- `representative_samples[]` gains `object_id` + `sample_record` (real in an authorized engagement); example updated.

## Pipeline audit (nothing else blocks real-data proof, at the code level)

| Layer | Finding |
|---|---|
| `bob_http_scan` | returns the **full real body** by default (`response_mode: "full"`) — already captures records |
| `redaction.js` (on-disk) | strips **secrets only** (tokens/keys/passwords/auth) — never emails/names/DOBs |
| `validateNoSensitiveMaterial` (evidence gate) | rejects **secrets/auth/cookies only** — a `{email,name,dob}` sample passes |
| claims | hold real data in `response_evidence` / `proof_of_concept` |
| report-writer / reporting.md | no forced redaction |

**Deliberately NOT touched (legitimate guards):** secret/token/cookie/auth redaction, RPC-endpoint redaction (protects operator infra), the brutalist sending only a *redacted* excerpt to the external roast LLM, and the ≤10-sample / no-full-corpus cap.

## Pairs with

PR #140 (over-permissive-search entry) pulls the sample; this lets the evidence pack keep it. Two halves of one feature: *Bob demonstrates data-exposure impact with a real sample, gated on authorization.*

## Tests

Regenerated claude/codex/kimi adapters. `test:mcp` 2891/0, `test:install` 15/0, `test:prompts` 193/0, `check:syntax` clean.

@codex review
@coderabbitai review


vmihalis:
Proactively applied the same **P1** Codex raised on #140: removed the process-global `BOB_DATA_ACCESS_AUTHORIZED` env flag (left set from a client run it would surface unmasked records on a later public safe-harbor target). The masked-vs-unmasked decision now keys on the session's loaded program policy — session-scoped, default-safe. (commit 1c49865)

@codex review
@coderabbitai review

vmihalis:
Closing as **redundant** — after tracing the actual proof path, this PR adds a leak surface without adding any client-facing value.

**Why it's redundant.** A web finding's report proof comes from the finding's own `proof_of_concept` / `response_evidence` (report-writer renders the "PoC" and "Evidence (response proving the bug)" lines from those fields). Those are free-text, validated for *secrets only*, so the real 5–10 record sample passes through. That path is driven entirely by **#140**: the evaluator pulls the sample and records it in `response_evidence` → it lands in the client report unmasked. The evidence-pack `representative_samples` this PR modified is a *separate* grade-gating artifact that the grader reads (needs only field names + count for severity) — it is **not** what the client report quotes.

**Why it's a net negative.** Putting real PII into the evidence pack — as the round of P1/HIGH reviews here unanimously flagged — turns a bounded grade-gating artifact into a sensitive-data sink: `bob_read_evidence_packs` is `sensitive_output: false`, the markdown mirror (`evidence-packs.md`) writes samples verbatim and **syncs to the operator's VPS**, and the ≤10 cap isn't enforced on `sample_record` content. Doing it safely would need real code work (sensitive readback, session-bound authorization, enforced record cap) — all to make a *redundant* copy safe.

So: the offensive-proof goal (real records in the client report) ships with **#140**; the evidence pack stays masked (correct — it doesn't need the PII). No code change needed here.

PR #146, opened 2026-06-20
Bind lab-target authorization to an operator env var, not a published token

ejmockler:
## Security fix: lab-target self-authorization

The `bob_init_session` `lab_authorization` escape — the only sanctioned way past the public-DNS scope gate, for operator-owned loopback/RFC1918 lab hosts — was gated by a hard-coded ack token (`i-own-and-am-authorized-to-test-these-private-targets`) that was **published verbatim in the tool schema**. A model that can call `bob_init_session` could read the token from its own schema, self-construct `lab_authorization`, and scope a session to a private IP (e.g. `192.168.1.53`, cloud-internal ranges) with **no operator action** — a self-authorization hole.

## Fix

The attestation now requires `BOB_LAB_TARGET_ACK` (set to the exact token) in the **MCP server's environment** — an operator-only control, since a running model cannot mutate the server process's environment even with shell access in its own process. Specifically:

- The tool argument carries only the **non-secret intent** (`private_targets: true`); the `ack` field and the token are **removed from the `bob_init_session` schema** (`additionalProperties: false`).
- `parseLabAuthorization` reads the env gate **before** any attacker-influenced field of the argument, so a hostile in-process getter can't set the env as a read side-effect.
- All three lab entry points (`assertHttpScopeDomain`, `validateHttpScanScope`, `labTargetPermitted`) route through this single env-gated chokepoint — no ungated path. The persisted `lab-authorization.json` remains audit-graded (write-guarded) and is re-validated against the env on every read.

## Operator path (unchanged capability, new control)

To scan a private lab host you own: `export BOB_LAB_TARGET_ACK=i-own-and-am-authorized-to-test-these-private-targets` before launching the MCP server, then pass `lab_authorization: { "private_targets": true }` to `bob_init_session`. Documented in `docs/FIRST_RUN.md`. Eligibility stays narrow (IPv4 loopback + RFC1918 only; metadata/link-local/IPv6/`.internal` never eligible), scope is pinned to the exact host, and the default (direct) egress profile is required.

## Verification

- **Adversarial red-team** across five channels (argument, persisted-file, chokepoint-bypass, env-gate integrity, eligibility/SSRF) — **no bypass** under the threat model (agent controls all tool arguments/JSON but cannot set the server env).
- Direct: env **unset** + the published token in the argument → rejected (`not a public DNS domain`); env **set** → accepted.
- Tests parameterized over the env states, incl. the getter-side-effect case and the no-env / wrong-env / forged-file paths. Full `npm test` green locally.

### DIFF
--- a/docs/FIRST_RUN.md
+++ b/docs/FIRST_RUN.md
@@ -117,17 +117,18 @@ Do not use a real company, public service, customer environment, or bug bounty t
 
 ### Private / lab targets (loopback & RFC1918)
 
-By default Bob's scope kernel rejects any non-public target — bare IPs, loopback, and RFC1918 hosts all fail with "not a public DNS domain", because a registrable public domain is the ownership signal Bob relies on. To scan a private lab host that **you own and are authorized to test**, the **operator** must enable the escape out-of-band; the agent cannot enable it for itself:
+By default Bob's scope kernel rejects any non-public target — bare IPs, loopback, and RFC1918 hosts all fail with "not a public DNS domain", because a registrable public domain is the ownership signal Bob relies on. To scan a private lab host that **you own and are authorized to test**, the **operator** must enable the escape out-of-band; the agent cannot enable it for itself (it controls tool arguments, but never the server's environment):
 
-1. Set the operator attestation in the MCP server's environment **before launch** (the agent cannot read or set the server's environment, so this is an operator-only control):
+1. Set **two** operator controls in the MCP server's environment, and keep them set for the whole session (they are re-checked at every scan, not only at init — if you relaunch the server without them, scans of the lab target start failing with "not a public DNS domain"):
 
    ```text
    export BOB_LAB_TARGET_ACK=i-own-and-am-authorized-to-test-these-private-targets
+   export BOB_LAB_TARGET=192.168.1.53        # the exact host(s) you authorize; comma-separated for several
    ```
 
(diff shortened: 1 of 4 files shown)


ejmockler:
Addressed the follow-up review findings:

- **Bind lab ack to the requested host / process-wide grant (P1)** — fixed in `013d1e5`: added a second operator-only control, `BOB_LAB_TARGET` (the exact authorized host(s), comma-separated). `labTargetPermitted` now requires the host to be in that set, so an active ack cannot be pivoted to a neighboring private host. Verified: with `BOB_LAB_TARGET=192.168.1.53`, `192.168.1.53` is accepted but `10.0.0.9`/`192.168.1.99` are rejected; ack set + `BOB_LAB_TARGET` unset grants nothing.
- **Docs omit "every launch" (medium)** — fixed in `013d1e5` (`FIRST_RUN.md`): both env vars must stay set for the whole session (the gate is re-checked at scan time, not only at init), with the failure mode spelled out.
- **Audit artifact omits the authorized host (medium)** — fixed in `d26e85a`: `lab-authorization.json` now records `target_host` and `authorized_hosts` (the `BOB_LAB_TARGET` set at mint time), so the artifact is self-describing post-incident. The live gate still reads the env, not the file.

Tests: added the neighbor-pivot rejection case and the audit-record assertions; full `npm test` green.

PR #150, opened 2026-06-21
feat(evaluate): map primitives→signed producers in the evaluator role (P12)

vmihalis:
## What & why

The evaluator agent is **granted** the signed offensive producers as tools
(`bob_http_cors_confirm`, `bob_http_xss_reflect`, `bob_http_xss_confirm`,
`bob_oob_mint`/`bob_oob_poll`) but the role body **never told it when to call
them**. On the last full authorized run they minted **0 signed rows across 716
tool calls** — the machinery was starved purely by missing prose, not by code.

This adds a `Signed offensive producers` block to the evaluator role mapping
each proven primitive to its producer, plus the step that turns a minted row
into a credited finding.

## The mapping

| Primitive | Producer(s) | Ceiling |
|---|---|---|
| Reflected XSS (sink discovery) | `bob_http_xss_reflect` | MEDIUM |
| Reflected XSS (proven execution) | `bob_http_xss_confirm` (drives Bob's headless browser, **WAF-capable**) | HIGH |
| Credentialed cross-origin read | `bob_http_cors_confirm` | MEDIUM |
| Blind OOB injection (SSRF/RCE/XXE/blind SQLi) | `bob_oob_mint` → `bob_oob_poll` | (infra-gated) |

Plus a **"recording a mint"** step: a signed row is credited only when a claim
cites it, so on `row_written: true` the agent must record the finding with an
`evidence_refs[{ kind: "exploit_run", run_id, … }]` built straight from the
producer's return, at the row's stamped `demonstrated_severity` (a higher claim
severity is rejected; an uncited row earns no credit).

## Honest notes baked into the prose
- `bob_http_idor_confirm` is **INERT** (`blocked_by_design`, pending PR-D
  synthetic-identity provisioning) — the prose tells the agent not to depend on
  it and to prove IDOR via the existing `auth_profile="victim"` replay instead.
- OOB is **infra-gated** (needs `BOB_OOB_HOST`); on `blocked_by_infra` record a
  `blocked_prereqs[]` / `blocked_harness_runs[]` entry.
- These producers prove the **reportable** forms of CORS/XSS (vs the
  never-record baselines a few lines below), and **server-mint their own
  payloads/origins/canaries** — the agent passes only selectors (param ordinal,
  surface_id), never raw payloads or operator data.

## Scope / safety
- **Prose-only.** No runtime/JS change. Edits `prompts/roles/evaluator.md`
  and the regenerated `claude` / `kimi` / `codex` surfaces.
[...]

Issue #26, opened 2026-05-10
F3: Findings-index health pipeline event


vmihalis:
Priority: 3 of 6

Source: post-v1.3.0 follow-up after PR #21 / PR #23.

Scope:
- Add `safeAppendPipelineEventDirect("finding_index_failed", ...)` on the catch path in `findings.js:recordFinding`.
- Surface this health signal in `/bob-status` and `/bob-debug`.

Acceptance notes:
- Do not block finding recording on index-health telemetry failures.
- Add focused coverage for the failure event and status/debug rendering path.

vmihalis:
Implemented in the current tree; tests cover the finding_index_failed event and status/debug rendering path.

Issue #25, opened 2026-05-10
F2: Brief-slice registry


vmihalis:
Priority: 2 of 6

Source: post-v1.3.0 follow-up after PR #21 / PR #23.

Scope:
- Refactor `hunter-brief.js` to iterate a slice registry with per-slice budgets.
- Use the PR #21 brief-size test as the regression guard.

Acceptance notes:
- Preserve existing web and smart-contract brief contents.
- Make slice inclusion/budgeting explicit in the registry.
- Keep `test/hunter-brief-size.test.js` green and meaningful.

vmihalis:
Implemented in the current tree; hunter brief slice registry and brief-size regression test are already present.

Issue #24, opened 2026-05-10
F1: Capability registry tag


vmihalis:
Priority: 1 of 6

Source: post-v1.3.0 follow-up after PR #21 / PR #23.

Scope:
- Add `capability_id?` to the tool registry contract.
- Migrate `capability-metrics.js` and `capability-eval-harness.js` to derive capability maps from the registry instead of hand-listing.

Acceptance notes:
- Keep the registry contract generic.
- Preserve existing metrics/eval behavior while removing duplicate hand-maintained maps.
- Add focused tests for registry-derived mapping.

PR #151, opened 2026-06-21
feat(observability): X8 live in-run SSE dashboard route

vmihalis:
## What
Plane-Δ **X8** (steal-list **S8**) — first node of the "measured & demo-able spine" thrust. Adds a read-only Server-Sent-Events route `GET /api/session/:domain/events` that **upgrades the existing `/api/snapshot` poll** into a live, resumable stream, so the operator sees `wave_started → evaluator_stopped → wave_merged → finding_recorded → verification_written → grade_written` transitions **without manual refresh**.

Spec: `docs/plane-delta/detail/X8.md`.

## How
- **`mcp/lib/dashboard-event-tail.js` (new)** — tolerant fold of the two append-only ledgers (`frontier-events.jsonl` + `pipeline-events.jsonl`) into one `(ts, source_rank, recordId)`-ordered frame stream. Skips a partial trailing line (never throws, unlike `readJsonlStrict`); content-addressed frame ids (`FE-`/`PE-`); `Last-Event-ID` resume with trim-safe `resync`; 2 KB/frame budget; primitive-only compaction so no raw body / secret can leak.
- **`mcp/lib/dashboard.js`** — `openEventStream`/`writeFrame`/heartbeat helpers + the route (loopback-**hard-gated** 403 off-loopback; `?backlog=` cap; 1 s poll, 15 s heartbeat; timers `unref`'d and cleared on disconnect) + a client `EventSource` live feed (auto-subscribe to the most-active session, click a row to switch, debounced snapshot refresh on findings/wave/grade). `renderDashboardHtml` exported for the HTML test.

## Doctrine
- **Read-only (S2)** — writes nothing to `sessionDir`.
- **No MCP tool / role-bundle / generator / settings change** → 0 generated-surface drift, **0 added LLM tokens**.
- **X-rej-4** honored — file-backed tail over the existing ledgers, no in-memory bus.
- Loopback-only (unauthenticated live tail of session internals).

## Tests
- 4 SSE route tests in `test/dashboard.test.js` (fold streams frontier+pipeline; `Last-Event-ID` resume w/o replay; off-loopback → 403; survives a partial trailing line).
- 9 unit tests in `test/dashboard-event-tail.test.js` (ordering, id synthesis, tolerant parse, no-leak compaction, frame budget, resume/resync) — registered in `mcp-test-manifest.json`.
- Full `npm test` green locally.

## Summary by CodeRabbit

* **New Features**
  * Added a loopback-only SSE live tail endpoint for per-session event frames (GET + headers-only), with `Last-Event-ID` resume, periodic pings, and `resync` behavior; includes a new tolerant event-frame reader powering the stream and client “Live” updates.

* **Bug Fixes**
  * Improved robustness against missing/partial/malformed JSONL ledgers, tightened deterministic ordering, and enhanced sanitization/redaction plus per-frame size budgeting with safe truncation.

* **Tests**
  * Expanded coverage for tolerant ledger reading, cursor/resume edge cases, access restrictions, SSE parsing, and truncation behavior.

* **Chores**
  * Slightly increased npm package size budget.


vmihalis:
@codex review

vmihalis:
@coderabbitai review

vmihalis:
Addressed all review findings in `166fcf7`:

| # | Finding | Fix |
|---|---|---|
| 1 | **Scalar-secret leak** (CodeRabbit Major / Codex P2 / Brutalist) — `compactFrontierEvent` type-filtered, not key-allowlisted | Strict **key-allowlist** (`FRONTIER_PAYLOAD_KEY_ALLOWLIST`, structural/enum/numeric keys only); every surfaced string run through `redactTextSensitiveValues` + control-char-stripped; non-allowlisted scalars and all nested values dropped |
| 2 | **Char vs byte budget** (CodeRabbit) | `enforceFrameBudget` now uses `Buffer.byteLength(data, "utf8")` |
| 3 | **Invalid timestamps** (CodeRabbit / Codex) | Records whose `ts` doesn't parse (`timestampMs <= 0`) are rejected before key construction |
| 4 | **PE-hash dedup collision** (Brutalist) | Duplicate record ids within a read are disambiguated (`…~N`), so two same-content pipeline events are never silently deduped |
| 5 | **Caught-up resume re-send** (Codex) | `lastKey` is seeded from the resume cursor, so a reconnect at the newest frame doesn't re-send history |
| 6 | **SSE id injection** (Brutalist) | Frame-id components are control-char-stripped at construction (`stripControlChars`/`sanitizeIdComponent`) and CR/LF-stripped again in `writeSseFrame` |
| 7 | **O(N) poll** (Brutalist) | The poll skips the read+parse+sort when neither ledger's size/mtime changed (`sessionLedgerStamp`) |

**Tests:** +6 covering allowlist+redaction, byte budget, invalid-ts rejection, dedup disambiguation, id sanitization, and caught-up resume. Poll/heartbeat are now env-tunable (`BOB_SSE_POLL_MS` / `BOB_SSE_HEARTBEAT_MS`, production defaults unchanged) for deterministic fast teardown — which also fixed a 16 s test hang where `server.close()` waited on the heartbeat to detect a client-destroyed socket. Full `npm test` green.

vmihalis:
@codex review

vmihalis:
@coderabbitai review

vmihalis:
CI fix (test-only): the `oversized frames are truncated` test was Node-version-dependent — `redactTextSensitiveValues` flags the `x*200` filler as a secret on Node 20/22 (shrinking the payload below the cap) but not on Node 24, so truncation couldn't be forced deterministically through the redaction path. Replaced with a direct `enforceFrameBudget` unit test (no redaction) + an e2e invariant (every frame ≤ MAX_FRAME_BYTES regardless of redaction). No production logic change beyond exporting `enforceFrameBudget`.

vmihalis:
Round-2 review (Brutalist/Codex) — CodeRabbit reported **no actionable comments** 🎉; addressed the Brutalist/Codex findings as follows.

### Fixed
- **🟠 HIGH — `readJsonlTolerant` had no size cap (OOM):** now tail-reads at most `MAX_LEDGER_READ_BYTES` (8 MB) via an fd + `readSync`, dropping the partial first line. Bounds memory regardless of ledger size. *(+test: tail-read returns a bounded suffix, no partial record.)*
- **🟠 HIGH — pipeline events bypassed redaction (S1/S7 leak):** `normalizePipelineEventForRead` caps but doesn't redact freeform fields (`status`/`source`/`*_reason`). Added `redactPipelineEvent` — every pipeline string value now goes through `redactTextSensitiveValues` + control-char strip on the read path, matching the frontier treatment. *(+test: a `Bearer` token in `status` is redacted on the stream.)*
- **🟡 MED — raw fs error disclosed session paths:** the SSE catch now writes a generic `: tail-error` comment, never the error text.
- **🟡 MED — resume bypassed `SSE_MAX_BACKLOG`:** a resume flush is now capped to `SSE_MAX_BACKLOG`; an over-cap (very old) cursor elides the gap and emits a `resync`.

### Accepted as defensible tradeoffs (rationale, not fixed)
These are inherent to a **loopback, append-only, snapshot-backstopped** live tail; fixing them adds complexity disproportionate to the route's role:
- **Out-of-order poll drop:** ledgers are append-only and ts-monotonic in practice; a frame with a ts below the high-water mark is pathological, and `/api/snapshot` always reflects full state. Catching reordering would need unbounded sent-id tracking.
- **`~N` duplicate-id suffix is read-position-dependent:** pipeline events have no native id; content-hash + occurrence-ordinal is the pragmatic stable id. A front-trim that splits a duplicate run shifts ordinals — already signalled by the existing trim→`resync` path.
- **O(N) read when the ledger actually changes:** the `sessionLedgerStamp` skip eliminates the idle case (the common one); a re-read on real change is inherent to a file-backed tail (X-rej-4 forbids an in-memory bus), and ledgers are now bounded (5k/20k records + 8 MB tail).
- **`statSync` non-ENOENT absorbed:** worst case the stamp is `0:0` and the next poll re-reads — fail-open to a read, never a crash.
- **`resync` has no `id:`:** `resync` only fires when surviving frames exist and all sort after the cursor, and those (id-bearing) frames are emitted immediately after, so `Last-Event-ID` advances — no reconnect loop.

Also: fixed the Node-version-flaky budget test (direct `enforceFrameBudget` check, no redaction dependency). Full `npm test` green (27 dashboard/tail tests).

vmihalis:
@codex review

vmihalis:
@coderabbitai review

vmihalis:
## Round-4 convergence — `d699980`

Three correctness/robustness fixes addressing the remaining round-3 edge cases. All 9 lines; full `npm test` green locally.

**1. Tail — reject present-but-unparseable pipeline `ts`** (`dashboard-event-tail.js`)
The frontier branch already rejected invalid timestamps, but the pipeline branch let `normalizePipelineEventForRead` silently substitute `now()` for a malformed `ts` — fabricating frame ordering from corrupt input. Now a present-but-unparseable `ts` drops the record; an absent `ts` is still allowed (→ now), matching the frontier-branch guard.

**2. Dashboard — capture change-stamp _before_ the initial read** (`dashboard.js`)
`sessionLedgerStamp(domain)` was sampled _after_ `readSessionEventFrames`. An append landing between the read and the stamp would be captured by the stamp yet absent from the streamed frames, so the poll-skip optimization could mark the stream "caught up" and never flush it. Stamping before the read closes the window — the next poll always observes a changed stamp and flushes.

**3. Test — non-vacuous caught-up-resume assertion** (`dashboard.test.js`)
Added `statusCode === 200` + `ids.length >= 1` preconditions on first connect and `statusCode === 200` on resume, so `assert.deepEqual(resumedIds, [])` can't pass for the wrong reason (e.g. a failed connection that simply returned zero frames).

No new tool, role-bundle, generator, or audit-graded-path surface — still a read-only loopback-only SSE route folding the two on-disk ledgers.

@coderabbitai review
@codex review

vmihalis:
## Round-5 — disposition of all open Codex findings

CodeRabbit (the App gate) verified the round-4 fixes clean ("No actionable issues remaining"). Working through the Codex advisory findings from all three reviews; here is the disposition of each, with the two remaining real ones fixed and the one design-bounded item documented.

| Codex finding | Disposition |
|---|---|
| `dashboard.js:715` — mark unseen appends as caught up | ✅ Fixed round-4 (`d699980`) — stamp captured before read |
| `dashboard-event-tail.js:204` — reject invalid pipeline `ts` | ✅ Fixed round-4 (`d699980`) — present-but-unparseable `ts` dropped |
| `dashboard-event-tail.js:213` — redact explicit frontier IDs used as SSE IDs | ✅ **Fixed (round-5)** |
| `dashboard-event-tail.js:301` — signal resync when one ledger trims the cursor | ✅ **Fixed (round-5)** |
| `dashboard.js:730` — same-ts/same-source frame can be dropped | 📝 **Documented tradeoff (round-5)** |

### Fixed — `:213` redact explicit frontier event IDs (S1/S7)
`bob_append_frontier_event` accepts explicit `event_id`s; `compactFrontierEvent` redacted the payload copy, but the frame `id:` line (= the Last-Event-ID cursor) embedded the raw value with only control-char stripping. A secret-shaped id (`FE-…Bearer sk-…`) would reach the unauthenticated wire. `sanitizeIdComponent` now runs the repo redactor too. It's deterministic, so the cursor still round-trips for resume. New test: *"a secret-shaped explicit frontier event_id is redacted in the frame id, not just the payload."*

### Fixed — `:301` per-source resync detection
The old `after.length === frames.length` check was defeated by any surviving *older* frame in the **peer** ledger — masking a real front-trim gap in the cursor's own source (the two ledgers trim independently). `framesAfter` now checks per-source: resync fires when the cursor is gone **and** every surviving frame *from the cursor's own source* sorts after it. New test reproduces Codex's exact scenario (frontier trimmed past the cursor while pipeline retains an older frame) — resync now correctly signals.

### Documented — `:730` same-ts/same-source ordering
Real but design-bounded. The total order is `(ts_ms, source_rank, record_id)` where `record_id` is a **content hash, not append order**. So two frames sharing an exact millisecond *and* source order by hash, and the live poll's high-water-mark can skip a late-appended frame that hashes lower than the last one emitted.

Why documented rather than fixed:
- **Impact is bounded and self-healing.** The live feed is a best-effort 25-line ticker; the authoritative aggregates come from `/api/snapshot`, which the client auto-refreshes on significant events. A skipped frame still renders on any full backlog read (fresh connect / refresh). No data loss — the ledgers are the source of truth.
- **A complete fix is out of model.** The Last-Event-ID resume carries a *single* cursor, so the resume path can't both avoid re-sending and avoid dropping same-ms items without server-side per-event sequence numbers.
- **The alternative breaks a core invariant.** Append-position cursors would order correctly but shift under the front-trimming both ledgers do at their max-records cap — exactly what the content-addressed, trim-safe cursor model was chosen to avoid.

Documented at `compareFrameKeys` and at the poll call site so the tradeoff is explicit for future readers.

Full `npm test` green; `dashboard-event-tail` + `dashboard` suites 29/29.

@coderabbitai review
@codex review

vmihalis:
## Round-6 — CI pack-size budget (`c4ad0ec`)

CI failed `package.test.js` with `npm pack size 3300833 exceeds 3.3 MB` — but the **parallel CI run on the same commit packed under 3.3 MB**, and locally it packs at 3,297,518 B. So the budget was flapping at ~0 KB stable headroom (npm-pack tarball size is non-deterministic by ~3 KB across runs/envs).

All content-policy assertions pass (no `test/`, `node_modules/`, or internal docs leaked) — purely the size budget, which the X8 SSE observability runtime (`dashboard-event-tail.js` + the `dashboard.js` SSE route/client) legitimately grew. Following the documented convention in that test (3.1 → 3.2 → 3.3 MB bumps "without trimming assets"), raised to **3.4 MB** to restore a stable margin.

(Aside, not in scope here: `docs/hacker-bob-social.png` is 921 KB of the packed tarball — a candidate for a future asset-trim PR if the npm package shouldn't ship social-card images.)

@coderabbitai review
@codex review

vmihalis:
## Merged `main` (#150) into the branch — `196b434`

#150 landed on `main` while this PR was open and touched `test/package.test.js`, so the branch went `DIRTY`. Merged `main` in:

- **Conflict** was only the `package.test.js` pack-budget comment block — **both** #150 and this PR independently raised the assertion to the **identical 3.4 MB** value. Resolved by merging the two justifications (the bump is honestly credited to *both* #150's offensive-surface growth and X8's observability runtime).
- The merge also brings in **`scripts/release-check.js` at 3.4 MB** (from #150) — this is the second, independent pack-size guard that caused the earlier CI failure (`canonical pack size 3300833 exceeds 3.3 MB`). With both guards now at 3.4 MB, that failure is resolved.

### ⚠️ Pre-existing flaky test disclosure (NOT introduced by this PR)
The post-merge full suite surfaced one intermittent failure, unrelated to X8:

```
✖ poll DNS-only hit: a sufficient MEDIUM positive flagged dns_only_attribution_weak
   AssertionError: { offensive_outcome: "blocked_operator_pii", reason: "capture_contains_sensitive_value" }
```

**Root cause:** `oob-collector.js` screens the bounded capture via `sensitiveShapesPresent()` (`offensive-http-common.js`), which runs over content containing the **minted OOB token**. `oobMint` occasionally produces a random token that trips a `SECRET_SHAPE_RES`/PII regex → a false `blocked_operator_pii`. It passes 19/19 in isolation (3×) and a `test:mcp` re-run went green; observed ~1/6 in the full suite.

This is a latent bug in the **offensive OOB subsystem** (untouched by both X8 and #150) — a minted nonce should be guaranteed not to match the secret/PII screen. I'm **not** fixing it in this PR: it's a different, security-sensitive subsystem (the screen is a PII guard) and belongs in its own focused, reviewed change. Flagging for a follow-up de-flake PR.

X8 code is unchanged since the round-5 review (CodeRabbit clean); this push is the main-merge + the resolved budget conflict.

@coderabbitai review
@codex review

vmihalis:
## Final thread disposition (pre-merge)

CodeRabbit (the authoritative App gate) is clean across all six rounds + the merge commit. The 43 open threads are Codex/Brutalist advisory re-posts accumulated across 6 commits (most reference superseded code). Logging the disposition of each remaining family before resolving — per the merge rule, every finding gets a call on the record:

**Fixed in current code:**
- *Pipeline invalid-`ts` before normalization* (`dashboard-event-tail.js:209/210`) → guard at line 202 drops present-but-unparseable `ts` (round-4).
- *Redact explicit frontier IDs used as SSE IDs* (`:219`) → `sanitizeIdComponent` now redacts (round-5).
- *Pipeline SSE serves unredacted freeform fields* — **HIGH** (`:204`) and *don't stream unvalidated pipeline strings* (`:165`) → `redactPipelineEvent` redacts every string field on the read path (round-2/5).

**Documented / defensible tradeoffs** (advisory; all share one root — the content-addressed, front-trim-safe `(ts, source, record_id)` cursor, chosen so resume needs no server-side state):
- *O(N) read+parse+sort per poll tick* (`dashboard.js:733`, ×3) — gated by the size/mtime stamp (no read when nothing changed) and bounded by each ledger's `MAX_RECORDS` cap; loopback-only, read-only, 1 s cadence. The alternative (byte-offset incremental tailing) breaks the trim-safe cursor. Accepted.
- *Out-of-order / same-ms / late-arriving frame drop* (`dashboard.js:734`, ×2) — documented at `compareFrameKeys` + the poll site; can't be perfectly ordered without per-event sequence numbers; self-heals on any full backlog read (snapshot is authoritative).
- *`~N` dup-id suffix unstable across front-trim* (`dashboard-event-tail.js:222`) — same family: the occurrence-disambiguator is read-relative and can't be made trim-stable without sequence numbers; impact is a reconnecting client missing a duplicate-content event, self-healing on full load.
- *Non-ENOENT `statSync` swallowed to `0:0`* (`:300`) — intentional fail-soft for a read-only tail, consistent with the deliberate no-fs-path-leak doctrine (the generic `tail-error` marker); a persistent EACCES degrades to "empty stream" rather than leaking fs detail.

No remaining finding requires a code change. Resolving the threads and merging.

(@coderabbitai — yes, please open the two follow-ups you offered: the `oobMint`/`sensitiveShapesPresent` de-flake, and the `docs/hacker-bob-social.png` npm-pack exclusion.)

vmihalis:
## Round-7 — brutalist-surfaced findings on the merge head (`862d2f2`)

The brutalist run on the merge commit surfaced findings CodeRabbit had passed clean. Read each; the legitimate ones are fixed (all in X8's own surface), the rest dispositioned.

**Fixed:**

| Finding | Fix |
|---|---|
| **P1 — DNS-rebinding** (`dashboard.js`): SSE loopback gate checked only the bind address, so a rebound `evil.example→127.0.0.1` page becomes same-origin and could read live session frames | `requestHostIsLoopback()` (URL-parsed, port/IPv6-safe, **fail-closed**) now required on the SSE **and** snapshot routes when bound to loopback; remote-bind mode unaffected. +2 tests |
| **Shallow redaction** (`dashboard-event-tail.js:162`): nested `counts{}` keys bypassed the redactor, breaking the module's "no unredacted string on the wire" invariant | `redactPipelineEvent` now recurses objects/arrays and redacts keys too. +1 test |
| **Frontier domain affinity** (`:195`): frontier path didn't reject cross-`target_domain` rows the pipeline path rejects | mirrored the affinity check (defense-in-depth). +1 test |
| **No SSE connection cap** (`dashboard.js:723`): N connections × full-ledger reads | `SSE_MAX_CONNECTIONS=64` (env-tunable), decrement-once release on teardown. +1 test |

**Dispositioned (defensible, logged):**
- *Out-of-order / late-arriving drop* (`dashboard.js:734`) — re-statement of the documented same-ms ordering tradeoff (content-addressed trim-safe cursor; self-heals on full load; snapshot authoritative).
- *Honor requested backlog on resume* (`dashboard.js:707`) — kept as-is: Last-Event-ID resume should deliver what the client *missed* (the EventSource resume contract), bounded by `SSE_MAX_BACKLOG` with a `resync` signal on truncation. Capping resume at the display-size `backlog` hint would silently drop events. No flood risk (the client renders only the last 25 lines) and no cross-trust exposure (loopback, operator's own session).

Full `npm test` green; `dashboard` + `dashboard-event-tail` suites 34/34.

@coderabbitai review
@codex review

vmihalis:
## Round-8 — brutalist/codex findings on `862d2f2` (`879b1f3`)

All genuine, all in X8's own surface. Note the 7 *re-posts* of already-fixed/documented items (`:828` DNS-rebinding, `:181` shallow redaction, `:218` frontier affinity, `:765` conn cap — fixed in round-7; `:776`/`:749`/`:245` — documented) are the bots re-emitting on the same line numbers; the current code already addresses them (tests prove it). The **4 new, legitimate** findings are fixed:

| Finding | Fix |
|---|---|
| **P1 — bare secret tokens not masked** (`tail:85`): `redactTextSensitiveValues` masks URL/`Authorization`/assignment *forms* only, not bare `ghp_`/`AKIA`/`AIza`/`xox`/JWT/PEM. Pipeline writes validate only 2 freeform fields, so a bare token in `status`/`source` could stream | read path now also applies the canonical `SENSITIVE_VALUE_RE` patterns (`scrub()`), **plus** a frame-level fail-safe in `enforceFrameBudget` that stubs any frame still matching a secret shape (defense-in-depth for non-string/overlooked paths). +2 tests |
| **P2 — slot leak** (`dashboard.js:794`): close/error cleanup was registered *after* the catch-up read, so a mid-read disconnect leaked the round-7 connection-cap slot + timers | register teardown **before** the read; `cleanup` no-ops on null timers |
| **P2 — absent-`ts` replay** (`tail:233`): a missing `ts` is filled with `now()`, re-hashing the `PE-` id every read → replayed each poll | require a valid *persisted* `ts` (drop absent too, not just present-but-unparseable), matching the frontier path. +1 test |
| **P2 — symlink/FIFO ledger** (`tail:94`): `openSync` followed symlinks / could block on a FIFO | `lstat` regular-file check before open (+ post-open `fstat` TOCTOU re-check). +1 test |

**Honest note for reviewers/maintainer:** this is round 8, and the adversarial reviewers have now surfaced **two P1s in two rounds** (DNS-rebinding, then this secret leak) — both now closed. An unauthenticated loopback route streaming session internals is an inherently sensitive surface; the trend is toward narrower P2 edge cases. If this round's review is clean (or only re-states the documented tradeoffs), this is the converge point.

Full `npm test` green; `dashboard` + `dashboard-event-tail` 38/38.

@coderabbitai review
@codex review

vmihalis:
@codex review

(Re-requesting a review specifically on the round-8 commit `879b1f3` — the prior Codex review is still pinned to `862d2f2`, which predates the bare-secret / slot-leak / absent-ts / symlink fixes.)

vmihalis:
## Round-9 — pipeline key allowlist (root-cause redesign) + 5 hardening fixes (`aad3a8c`)

The operator chose the **structural redesign** over continued patching, to end the recurring redaction-completeness P1 arc.

### Root cause — eliminated by construction
The pipeline read path now surfaces **only allowlisted** structural/enum/numeric/boolean keys (`PIPELINE_KEY_ALLOWLIST`), mirroring the frontier path. Freeform text (`status`, `source`, `*_reason`, `agent`) and identity metadata (`started_by`, `egress_*`) are **dropped**, not redacted — so the wire cannot carry an operator-pasted secret *regardless of redaction completeness*. The ticker's display fields (`type`, `ts`, `kind`, `surface_id`) are all retained.

### The 5 non-redaction fixes
| Finding | Fix |
|---|---|
| **P1 — marker-only redaction left the body** (`tail:83`) | `scrub()` redacts the **whole value** on any `SENSITIVE_VALUE_RE` match (PEM/Cookie bodies no longer survive). Belt behind the allowlist; also covers frontier fields + id components |
| **P1 — `openSync` follows symlinks (TOCTOU)** (`tail:133`) | open `O_RDONLY\|O_NOFOLLOW\|O_NONBLOCK` → swapped-in symlink throws ELOOP, swapped-in FIFO returns a fd (fstat rejects) instead of blocking |
| stamp advanced before a possibly-throwing read (`dash:783`) | poll advances the change-stamp only after a **successful** read+flush; failed initial read nulls the stamp to force a retry |
| `sessionLedgerStamp` used `statSync` vs reader's `lstatSync` (`tail:364`) | now `lstatSync` — consistent no-follow posture |
| module-global connection counter (`dash:42`) | per-server `sseState` threaded through `startDashboardServer → routeDashboardRequest → startSseEventStream` |

**Dispositioned:** cross-connection read dedup (`dash:788`, perf) — bounded by the per-server 64-connection cap + size/mtime stamp-skip; an optimization, not a correctness/security issue.

Tests updated for the allowlist (freeform **dropped**, verified) + new whole-value PEM-redact and bare-token tests. Full `npm test` green; `dashboard` + `dashboard-event-tail` 39/39.

With freeform exposure gone structurally, the redaction-P1 class is closed at the source. Final review pass requested.

@coderabbitai review
@codex review

vmihalis:
## Round-10 — finish the no-freeform allowlist + cap default (`51e1a0f`)

The round-9 redesign produced **no new redaction P1** — only these 3 narrow *consistency completions* (all P2) of the allowlist, now fixed:

- **frontier `actor` dropped** (`tail:182`) — `compactFrontierEvent` still streamed the top-level freeform/identity `actor`; dropped to match the pipeline path (structural ids + `event_hash` remain).
- **`counts` dropped from the pipeline allowlist** (`tail:69`) — its keys are arbitrary, un-enumerated labels (not ticker display fields), so they don't belong in a strict structural allowlist; dropping closes the last freeform-key vector *by construction* rather than scrubbing.
- **`sseState` is now required** (`dash:805`) — removed the per-call `{ active: 0 }` default that would silently defeat the connection cap if a caller omitted it; `startDashboardServer` is the sole caller and always threads its per-server state.

**Dispositioned** (`tail:289` — "stop hashing dropped data into PE- ids"): a truncated SHA-256 over the full normalized projection is one-way (no secret recoverable from 96 bits of hash, and only with knowledge of every other field) and gives *better* event distinction — hashing only the allowlisted subset would increase `~N` dup-id collisions. Kept as-is.

Stale re-posts in the thread list (DNS-rebinding, O_NOFOLLOW, shallow-redaction, frontier-affinity, absent-ts, stamp-after-read, cleanup-order, module-global counter) are all fixed in current code — the bots re-anchor them each run.

+1 test (frontier `actor` dropped); counts test asserts drop. Full `npm test` green; `dashboard` + `dashboard-event-tail` 40/40.

Freeform exposure is now gone from **both** ledger projections. Final review pass.

@coderabbitai review
@codex review

vmihalis:
## Round-11 — shape-gate the last frontier freeform residuals (`8ce6d94`)

Round-10's review produced **no new P1** — the only genuinely-new findings were two **P2** consistency residuals of the freeform class on the *frontier* path (the pipeline path was already strict). This round closes them **by construction**, finishing the round-9/10 "eliminate freeform, don't rely on redaction" redesign symmetrically across both projections.

- **#24 (P2) frontier `tags[]`** (`tail:188`) — `tags` are caller-supplied via `bob_append_frontier_event` (normalized, not a closed enum). Now **shape-gated per tag**: structural labels (`money-movement`, `auth`) reach the wire; freeform/identity tags and pasted prose are dropped.
- **#26 (P2) frontier `payload.*` string values** (`tail:199`) — the keys were allowlisted but the string *values* were only scrubbed, so a freeform note in `payload.status`/`decision`/etc. still streamed. Now **shape-gated**: only structural tokens (enums/ids/state-codes) reach the wire; freeform prose is dropped. Numbers/booleans pass through unchanged.

New helper `structuralTokenOrNull` — `scrub()` first (so a redacted secret stays `"REDACTED"`, a secret masked into a spaced form is dropped), then `STRUCTURAL_TOKEN_RE` (`^[\w.:/+-]{1,64}$`: word chars + `. : / + -`, len 1..64, no whitespace/freeform punctuation). Top-level fields are unchanged — a fixed structural-id/enum/hash allowlist; `event_id` must round-trip as the Last-Event-ID cursor so it stays `safeStr`.

**Already fixed (stale re-anchors):** #20 (drop frontier `actor`) and #19 (drop `counts`) were done in round-10; the bots re-anchor them each run.

**Dispositioned — kept (#25/#21/#10, P2, "make duplicate frame IDs stable across trims"):** for two **byte-identical** events at the **same millisecond** and **same source**, there is no content-derived stable disambiguator — they are indistinguishable by content, so the `~N` suffix is a best-effort tiebreaker that prevents single-read dedup. A truly append-stable disambiguator needs per-event sequence numbers, which would (a) require schema changes to the two upstream producers (`frontier-events`, `pipeline-events`) — out of scope for a read-only observability overlay — and (b) not eliminate resync anyway, since these ledgers are front-trimmed (a seq cursor can still fall off the front, which `framesAfter` already handles). The live ticker self-heals on any full backlog read; this is the documented, accepted limitation of the content-addressed, front-trim-safe cursor model (see the `compareFrameKeys` header). Kept as-is.

**Net:** freeform content is now gone from **both** the frontier and pipeline projections by construction — every wire field is a structural id/enum/hash from a fixed allowlist or a shape-gated structural token. There is no freeform surface left to leak.

Tests: payload test now asserts a freeform scalar is dropped + a structural token kept; new `tags` shape-gate test; byte-budget integration test repurposed to assert freeform payload is stripped by construction. `dashboard` + `dashboard-event-tail` 41/41; full `npm test` green (3246/3246, 0 fail).

@coderabbitai review
@codex review

vmihalis:
## Round-12 — fix the shape-gate P1 + close the freeform class uniformly + 2 correctness P2s (`26c1bd7`)

Round-11's review found one **genuinely-new P1** (a real bug in the shape-gate I'd just added) plus same-class P2s on sibling fields and two separate correctness P2s. This round fixes the P1 and reaches the **fixed point** — every allowlisted string value across *both* projections is now shape-gated, with one documented exception (the cursor).

### P1 — `structuralTokenOrNull` leaked the prefix of overlong tokens (`tail`)
The helper did `safeStr(value, 64)`, which **truncated to 64 before** the regex ran — so an overlong all-token-char value (an opaque `sk_live_…`-style token that misses `SENSITIVE_VALUE_RE`) passed on its 64-char prefix. Fixed: **scrub without truncating, then reject any scrubbed value outside `[1, maxLen]`** before the char-class test. The length bound is now a *rejection* criterion, never a truncation. Regression test added (`LEAKPREFIX_` + 100 chars → fully dropped, no prefix on the wire).

### Freeform class closed uniformly (the field-by-field whack-a-mole, ended)
The gate was being applied one field at a time, so each review surfaced the next un-gated field. Now applied to **all** allowlisted string values:
- **frontier top-level ids** (`tail:205` P2) — `ts, kind, surface_id, frontier_item_id, task_id, claim_id, event_hash` shape-gated at `STRUCTURAL_ID_MAX=128`. **`event_id` is the one deliberate exception** — it's omitted from this projection and set authoritatively to the sanitized Last-Event-ID cursor in `buildFrames`, because it must round-trip arbitrary operator-supplied ids for resume (it's still control-stripped + secret-scrubbed there).
- **pipeline values** (`tail:68` P2, evidence-mode freeform `surface_id`) — shape-gated in `redactPipelineEvent`.
- **frontier tags + payload values** — already gated in round-11.
- Removed now-dead `safeStr` + `redactDeepValue` (everything routes through the one gate).

### Correctness P2s (dashboard)
- **`dash:763` — resume now honors `?backlog=`** (still capped at `SSE_MAX_BACKLOG`): a reconnect with `backlog=20` no longer flushes up to 1000 missed frames. `-0`-safe slice so `backlog=0` sends nothing + a `resync`. Server test added (backlog=2 over a larger gap → exactly 2 newest + resync).
- **`dash:573` — the client `resync` handler now reloads the snapshot** (`scheduleRefresh()`), so counters/rows reconcile across an elided gap (trim or backlog cap) instead of the live list just being cleared. Render-wiring assertion added.

**Net:** there is now no freeform string surface on the wire across either projection — every value is a structural id/enum/hash from a fixed allowlist, length-and-charclass gated, *or* the sanitized cursor. The recurring freeform class is closed by construction.

`dashboard-event-tail` 29/29, `dashboard` 16/16; full `npm test` green (3250/3250, 0 fail).

@coderabbitai review
@codex review

vmihalis:
## Round-13 — gate the `event_id` cursor + strengthen the secret regex + fix the backlog=0 cursor (`f6d48ab`)

Round-12 surfaced one more P1 plus a bug I'd just introduced. All three actionable items are fixed; the two remaining items are irreducible and dispositioned below — this is the convergence point.

### Fixed
- **`event_id` cursor bypassed the gate** (Brutalist medium + Codex P1) — it was the one caller-controlled string still scrub-only. `buildFrames` now runs an explicit frontier `event_id` through `structuralTokenOrNull`, falling back to the synthesized `FE-<hash>` when it isn't a structural token. A freeform/opaque/secret id never reaches the SSE `id:` line or `event.event_id`; the content hash still round-trips for Last-Event-ID resume; a structural explicit id is preserved as-is. (This is Brutalist's own suggested fix.)
- **bare tokens the scrubber missed** (Codex P1, e.g. `sk_live_…`) — added Stripe secret/restricted (`[sr]k_live/test_`; publishable `pk_` deliberately excluded — it's public) and OpenAI (`sk-`) patterns to the canonical `SENSITIVE_VALUE_RE`. These are enforced at **both** layers that consume it: per-value `scrub` *and* the `enforceFrameBudget` per-frame secret-shape fail-safe.
- **backlog=0 cursor advance** (Codex) — my round-12 cap set `toSend=[]` but left `lastKey` at the old cursor, so the poll re-emitted the elided frames on the next append. Now the high-water mark advances to the newest elided frame before the flush.

### Dispositioned — the irreducible residuals (this is the fixed point)
- **"a structural token could still be an opaque secret"** — after the above, the only remaining leak vector is a token that matches *no* secret pattern **and** is shaped *exactly* like a legitimate structural id (no whitespace, ≤128 chars, `[\w.:/+-]`). Such a value is by definition indistinguishable from a legit content-hash id / `event_hash` — which the dashboard must display. Closing it would require an entropy heuristic that false-positives on every legit hash id. It is dispositioned by the threat model: **loopback-only** (127.0.0.1, DNS-rebinding-guarded), **read-only**, serving Bob's **own** session telemetry to the operator's **own** browser — no external attacker. Brutalist itself rates this **medium, not high**, citing exactly this ("loopback-only/producer-supplied threat model … not attacker input"). The shape gate's job is to drop *freeform*; `scrub` + the fail-safe handle *known secrets*; this last sliver is accepted.
- **duplicate frame-id stability across trims** — content-addressed cursor model. A stable discriminator for two byte-identical same-ms rows needs a persisted per-event sequence number in the two upstream producers (`frontier-events`, `pipeline-events`) — out of scope for a read-only overlay, and it wouldn't remove resync anyway (front-trim still drops the cursor). Documented limitation (`compareFrameKeys` header); degenerate trigger. Kept.

**Net:** every wire field is now a structural id/enum/hash from a fixed allowlist, length-and-charclass shape-gated, scrubbed against an expanded secret definition, and screened by a per-frame fail-safe — including the cursor. The freeform/secret class is closed to its irreducible, threat-model-accepted core.

Tests: `event_id` non-token→hash + structural-id preserved; Stripe-token redaction; backlog=0 cursor-advance regression. `dashboard-event-tail` 31/31, `dashboard` 17/17; full `npm test` green (3253/3253, 0 fail).

@coderabbitai review
@codex review

vmihalis:
## Round-14 — fix the 4 genuine findings; disposition the localhost-perf items (`a056cad`)

Per the scope decision: fix the real correctness/security findings, disposition the localhost-perf + low-probability items. This is the merge round.

### Fixed
- **OpenAI `sk-proj-` redaction gap** (Codex P1) — round-13's `sk-` pattern missed hyphenated keys. Aligned the canonical `SENSITIVE_VALUE_RE` to the OpenAI shape the repo *already* uses in `offensive-http-common.js`: `sk-(?:proj-)?[A-Za-z0-9_-]{20,}`. Enforced at both `scrub` and the per-frame fail-safe.
- **symlinked session *directory*** (Codex P1) — `lstat` + `O_NOFOLLOW` only guard the final component, so a symlinked *parent* (the session dir itself) was still followed. `readJsonlTolerant` now `realpath`s the ledger's parent and refuses it unless it stays under the `realpath`'d sessions root.
- **orphan SSE poller on early disconnect** (Codex) — if a client aborted *during* the initial catch-up read, `cleanup` ran while `poll`/`heartbeat` were still null, then the timers were created and never cleared. `startSseEventStream` now tears them down if the slot was released during the read (`if (released) cleanup()`).
- **secret-shaped `event_id` cursor collision** (Codex) — such ids scrubbed to `"REDACTED"` and collapsed onto one cursor. They now fall back to the per-record `FE-<hash>`, so distinct secret ids get distinct, trim-stable cursors.

### Dispositioned (operator-approved)
- **per-connection ledger re-scan** (Codex rates HIGH *perf*) — real, but this is a **loopback-only** dashboard with realistically ~1 viewer; the read is **stamp-gated** (only on ledger change), each timer is `unref`'d, and concurrency is capped at 64. A proper fix is a shared per-stamp read cache — a non-trivial refactor of the poll path, tracked for a follow-up rather than blocking this read-only overlay.
- **change-stamp coarseness** (`size:mtimeMs`) — a same-size atomic rewrite landing inside the mtime-granularity window could be missed. Very low probability on modern (ns-mtime) filesystems, and append growth strictly increases size. A `ctime`/inode addition is the follow-up fix.
- **(carried) opaque-token-shaped-as-id + duplicate-frame-id trim stability** — the two irreducible residuals from round-13: a token matching no secret pattern *and* shaped exactly like a legit hash-id can't be separated without an entropy heuristic that false-positives on real ids; a fully trim-stable duplicate discriminator needs producer sequence numbers (out of scope for a read-only overlay). Both accepted under the loopback-only/read-only threat model (Brutalist rates this class medium, not high).

**Net:** every actionable correctness/security finding is fixed; the remaining items are localhost-perf / very-low-probability / irreducible, all dispositioned under the loopback-only, read-only threat model with follow-ups noted.

Tests: OpenAI `sk-proj-` redaction; symlinked-directory refusal; secret-id distinct-hash. `dashboard-event-tail` 34/34, `dashboard` 17/17; full `npm test` green (3256/3256, 0 fail).

@coderabbitai review
@codex review

Issue #131, opened 2026-06-19
Offensive sandbox: close the BEDROCK same-UID in-process signing-key forge boundary (UID/container/network separation)


vmihalis:
## The deferred boundary

The offensive signed-row contract (`offensive-runs.jsonl`, HMAC-signed, introduced in #108) defends against an agent confined to the **MCP tool surface + guarded-Bash surface**. It does **NOT** defend against an agent that executes arbitrary in-process code at the same UID — a `node -e` that `require()`s the 0600 handoff signing key (`mcp/lib/handoff-signing-key.js`) and hand-MACs a row around every gate. This is the **BEDROCK** boundary.

Every signed-row producer states this honestly and explicitly does not close it:
- #108 (proof contract), #119 (`bob_http_idor_confirm`), #128 (OOB collector), and #130 (PR-PROV — provenance stamping) all add zero key isolation.
- It is the same assumption Bob's wave-handoff signing already makes.

## What closes it (the sandbox PR)

Run the offensive runner + the row-signing under a **separate UID / container / network namespace** from the agent, so the signing key is unreadable by the agent's same-UID code and the egress path is a hard scope boundary. Scope sketch (to be refined):

- Signing key owned by a non-agent UID; the runner is the only process that can read it.
- Offensive container exec under UID/network separation (the wide-open runner from #123/#124/#125 hardened into a real isolation boundary).
- A scope-aware container egress proxy as the hard outbound boundary (relevant to the deferred nuclei/IDOR live arms).

## Until then

The signed rows are sound against the tool-surface threat model (which is what the producers are reviewed against), and the boundary is stated in every producer + this issue is the external tracking reference. Closing this is a prerequisite for *absolute* un-fakeability.

Refs: #108, #119, #123, #124, #125, #128, #130.

PR #155, opened 2026-06-22
feat(offensive): PR1 — trusted authed_fetch browser transport (WAF-capable)

vmihalis:
## PR1 — trusted `authed_fetch` browser transport (WAF-capable)

First of two PRs adding the offensive **mass-read** producer. This one adds the transport it needs; PR2 adds the producer. Design of record: `docs/OFFENSIVE_MASSREAD_PRODUCER_PLAN.md` (included here).

### Why
The offensive rail's in-process `safeFetch` is Cloudflare-blind — a bare-Node TLS fingerprint that CF 403s. That left the **broken-auth / mass-read** class unconfirmable on a signed rail (the sexjobs.nl crown-jewel F-2 was only provable through the real-Chrome channel). This adds a WAF-capable, authenticated transport.

### What
- **`mcp/browser-driver.js` — a trusted, server-side-only `authed_fetch` command.** Navigates to the target origin, then issues an **in-page `fetch()`** (real Chrome TLS/HTTP-2 → survives the WAF) and returns the body. Scope-checked (`assertSafeResolvedRequestUrl`), origin-drift guarded, wall-clock-raced (same pattern as `evaluate`), body capped at 2 MiB (`body_truncated` flag), **GET/POST read-intent only**.
- **Cookie auth injection at session start** (`context.addCookies`) supplied by the in-process producer; bearer/custom headers go per-request in `authed_fetch`.
- **`mcp/lib/browser-sessions.js`** — threads `authCookies` → `BOB_BROWSER_DRIVER_INIT.auth_cookies`.

### Security posture
- The **agent-facing `evaluate` sandbox is UNCHANGED** (`FORBIDDEN_EVAL_PATTERN` still blocks `fetch(`/XHR/WS). This transport builds the fetch expression *server-side* from a scope-checked URL and runs it via a path agents can't reach.
- **No `bob_browser_*` MCP tool maps to `authed_fetch`** — only in-process producer code (`sendCommand`) can invoke it. A regression test asserts no browser tool wrapper references it.
- Auth material flows **producer → driver, never agent → driver**.

### Tests (`test/browser-authed-fetch.test.js`, manifested)
- spawn-stub: `authCookies` threads into the init payload; null when unauthenticated (the control arm).
- source-level contracts (no Chromium): in-page fetch + `credentials:"include"`; scope-check + origin-drift + timeout-race + body-cap; method restriction; cookie injection; **sandbox-unchanged**; no agent tool exposes the command.
- **real-Chromium end-to-end differential** (via a `localtest.me` cookie-gated server, skips cleanly where that domain doesn't resolve to loopback): the authed arm reads the full collection (cookie carried by the in-page fetch); the no-cookie control is denied — the mass-read differential, proven for real.

Full `npm test` green (3262 pass, 0 fail; 2 pre-existing skips).

@coderabbitai review
@codex review

## Release Notes

* **New Features**
  * Added a trusted, cookie-scoped authenticated browser transport with strict URL/origin scoping, manual redirects, response-size caps, and timeout protection.
  * Injects auth cookies after the browser session is ready (one-time per session) via dedicated commands.
  * Unproxied browsing pins resolved targets to mitigate DNS rebinding risk, with fail-closed handling for internal hosts.

* **Documentation**
  * Added a design plan for differential validation of bulk PII-read authorization.

* **Tests**
  * Expanded contract and end-to-end coverage for scoping, truncation, header restrictions, and session behavior, including manifest updates.

### DIFF
--- a/docs/OFFENSIVE_MASSREAD_PRODUCER_PLAN.md
+++ b/docs/OFFENSIVE_MASSREAD_PRODUCER_PLAN.md
@@ -83,11 +83,17 @@ existing Patchright + system-Chrome stack (used by `bob_http_xss_confirm`) beats
   a wall-clock timeout race + an in-page `AbortSignal.timeout` (no lingering renderer
   request) + a body cap measured in **bytes** (not UTF-16 code units), enforced while
   streaming.
-- **DNS-rebinding pin** — the target host is resolved ONCE in Node and Chrome is launched
-  with `--host-resolver-rules=MAP <host> <validated-ip>`, so the browser connects to the IP
-  we checked (closing the TOCTOU where the static scope check passes but Chrome re-resolves
-  to an internal/attacker IP). Under `block_internal_hosts`, an unpinned host is refused
-  fail-closed. Skipped under an egress proxy (DNS resolves at the proxy).
+- **DNS-rebinding pin (required, not opt-in)** — the target host is resolved ONCE in Node and
+  Chrome is launched with `--host-resolver-rules=MAP <host> <validated-ip>`, so the browser
+  connects to the IP we checked (closing the TOCTOU where the static scope check passes but
+  Chrome would re-resolve to an internal/attacker IP). Because `authed_fetch` is ALWAYS
+  credentialed, the pin is REQUIRED on every call (not just under `block_internal_hosts`): an
+  unpinned host is refused fail-closed so a credentialed request can never ride a rebound
+  hostname. The pin also records whether the resolved IP is internal; under
+  `block_internal_hosts` a pinned-but-internal IP is refused too. **Producer contract:** start
+  the browser session with `target_url` on the exact host you will read, so that host is the
+  one pinned. Under an egress proxy the pin is N/A (DNS resolves at the proxy) and
+  `block_internal_hosts` is refused rather than silently downgraded.
 - **Refuse when `block_internal_hosts` is on** (same SSRF stance as `xss_confirm`, which
   cannot enforce the session SSRF policy through the browser) — the producer refuses; the
   driver's scope+pin checks are defense in depth for a direct call.
@@ -95,10 +101,13 @@ existing Patchright + system-Chrome stack (used by `bob_http_xss_confirm`) beats
   hostile target could in principle override `window.fetch`/`Response` to forge the result.
   This is inherent to the transport's purpose (real TLS fingerprint requires the page network
   stack), and a "native fetch capture" was tried and reverted — it is unreliable under the
-  Patchright stealth driver and broke the transport. Compensating controls: `waitUntil:
-  "commit"` minimizes page-script execution before the fetch, and the authed-vs-control
-  DIFFERENTIAL bounds integrity (page code cannot tell which arm it serves; a target forging
-  a vuln against itself is not a coherent threat). A single capture is evidence, not proof.
+  Patchright stealth driver and broke the transport. Compensating control: `waitUntil:
+  "commit"` minimizes page-script execution before the fetch. On integrity, the
+  authed-vs-control differential is NOT tamper-proof — the target CAN tell the authed arm (it
+  carries the cookie) from the control arm and could serve forged data — but a target forging
+  a vuln against ITSELF gains nothing (self-reporting), the per-tool ceiling bounds a forged
+  result, and a single capture is evidence, not proof (the operator corroborates for
+  integrity-sensitive use).
 
 ## Severity ceiling
 
@@ -147,7 +156,8 @@ The manual, fail-closed edits:
 - Auth material flows producer → driver over **stdin, never the env**; each cookie is
   scope-validated against `target_domain` before it reaches the browser context.
 - The browser is DNS-pinned to the Node-validated IP (`--host-resolver-rules`); an unpinned
-  host is refused under `block_internal_hosts`.
+  host is refused on EVERY (always-credentialed) `authed_fetch`, and a pinned-but-internal IP
+  is refused under `block_internal_hosts`.
 - `authed_fetch` does not follow redirects (`redirect:"manual"`), self-aborts on timeout,
   and caps the body in bytes.
 - Page-controlled fetch is an accepted residual (see transport section) — mitigated by
--- a/docs/OFFENSIVE_MASSREAD_PRODUCER_PLAN.md
+++ b/docs/OFFENSIVE_MASSREAD_PRODUCER_PLAN.md
@@ -160,8 +160,15 @@ The manual, fail-closed edits:
   is refused under `block_internal_hosts`.
 - `authed_fetch` does not follow redirects (`redirect:"manual"`), self-aborts on timeout,
   and caps the body in bytes.
+- The trusted `authed_fetch` op (priming nav + the credentialed fetch) is NOT recorded into
+  the agent-readable network log / record buffer — its cookies/headers never leak to agents.
 - Page-controlled fetch is an accepted residual (see transport section) — mitigated by
   `waitUntil:"commit"` + the differential design, not by an unreliable native-fetch capture.
+- Priming-redirect under `block_internal_hosts` is an accepted residual: a server-driven 3xx
+  during the priming nav could fire one request to an internal host before the drift guard.
+  It is NOT blocked with a catch-all network-route interceptor — that is forbidden (it aborts
+  Kasada/CDN/OAuth subresource loads, defeating the WAF-bypass purpose). Bounded by the
+  producer's refusal under `block_internal_hosts`, the pin, the drift guard, and the scope check.
 - Every request audited via `auditConfirmRequest`; refuse under `block_internal_hosts`.
 - `demonstrated_severity` stamped from the frozen registry, never agent-supplied; the
   per-tool ceiling bounds blast radius to a fabricated HIGH at worst (the #131 same-UID
--- a/docs/OFFENSIVE_MASSREAD_PRODUCER_PLAN.md
+++ b/docs/OFFENSIVE_MASSREAD_PRODUCER_PLAN.md
@@ -62,21 +62,43 @@ Two separate outputs, by design:
 ## The WAF transport (Chrome + a trusted `authed_fetch`)
 
 The in-process `safeFetch` is Cloudflare-blind (bare Node TLS fingerprint → 403). The
-existing Patchright + system-Chrome stack (used by `bob_http_xss_confirm`) beats CF. So:
+existing Patchright + system-Chrome stack (used by `bob_http_xss_confirm`) beats CF. So
+(implemented in PR1, hardened across review rounds 2–3):
 
-- **Auth injection at session start** (`mcp/browser-driver.js#start`): seed
-  `context.setExtraHTTPHeaders(...)` and/or `context.addCookies(...)` from the existing
-  `buildHeaderProfile` (`auth.js`). The browser has NO auth path today; this adds one,
-  threaded from the producer (never from an agent).
+- **Auth injection over stdin, never the env** — cookies are sent to the driver AFTER ready
+  via a server-side-only `set_auth_cookies` command (`mcp/lib/browser-sessions.js` →
+  `mcp/browser-driver.js#setAuthCookies`), so auth secrets never touch the process
+  environment (no child/renderer inheritance, no `/proc/<pid>/environ` snapshot). EACH
+  cookie's target host is scope-validated against `target_domain` before it reaches the
+  context — an out-of-scope cookie fails the whole session closed. Threaded from the
+  producer, never an agent.
 - **A trusted `authed_fetch` driver command** — server-side-only, NOT agent-reachable.
-  Issues a scope-checked authenticated request from the page context (`page.evaluate` of a
-  server-built fetch, or `page.request`/`context.request` with the injected auth) and
-  returns the body. The agent-facing `evaluate` sandbox (`FORBIDDEN_EVAL_PATTERN` — no
-  `fetch(`/XHR/WS) stays FULLY CLOSED; only the trusted producer path uses the new
-  command. Mirror the `xss_confirm` scaffolding: `assertSafeResolvedRequestUrl` +
-  read-only-path + final-URL drift guard + `auditConfirmRequest` on every request.
+  Issues a scope-checked authenticated request from the **page world** (`page.evaluate` of a
+  server-built `fetch`) — this is required to carry the real Chrome TLS/HTTP-2 fingerprint;
+  a Node-side `context.request` is CF-403'd like `safeFetch`. The agent-facing `evaluate`
+  sandbox (`FORBIDDEN_EVAL_PATTERN` — no `fetch(`/XHR/WS) stays FULLY CLOSED; only the
+  trusted producer path uses the new command. Scaffolding mirrors `xss_confirm`:
+  `assertSafeResolvedRequestUrl` (URL + origin) + `waitUntil:"commit"` (minimal authed-page
+  execution) + final-URL origin-drift guard + `redirect:"manual"` (no off-scope auth leak) +
+  a wall-clock timeout race + an in-page `AbortSignal.timeout` (no lingering renderer
+  request) + a body cap measured in **bytes** (not UTF-16 code units), enforced while
+  streaming.
+- **DNS-rebinding pin** — the target host is resolved ONCE in Node and Chrome is launched
+  with `--host-resolver-rules=MAP <host> <validated-ip>`, so the browser connects to the IP
+  we checked (closing the TOCTOU where the static scope check passes but Chrome re-resolves
+  to an internal/attacker IP). Under `block_internal_hosts`, an unpinned host is refused
+  fail-closed. Skipped under an egress proxy (DNS resolves at the proxy).
 - **Refuse when `block_internal_hosts` is on** (same SSRF stance as `xss_confirm`, which
-  cannot enforce the session SSRF policy through the browser).
+  cannot enforce the session SSRF policy through the browser) — the producer refuses; the
+  driver's scope+pin checks are defense in depth for a direct call.
+- **ACCEPTED RESIDUAL — page-controlled fetch.** The request runs in the page world, so a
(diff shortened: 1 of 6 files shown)


vmihalis:
## PR1 round-2 — all 7 transport-security findings fixed (`3b99bee`)

The isolated review did its job — every finding was a real issue in the new trusted transport. All addressed:

| Finding | Sev | Fix |
|---|---|---|
| `fetch` follows redirects → off-scope auth leak | P1 | `redirect:"manual"` — a 3xx surfaces as `type:"opaqueredirect"`/status 0 → producer's 2xx oracle fails closed |
| Body cap is post-hoc (`await text()` buffers all) | high | stream-read via `getReader()` with the 2 MiB cap enforced **while reading** + `cancel()` |
| Authenticated root-page load runs site JS | P1 | `waitUntil:"commit"` (minimal execution; same-origin context still required so a minimal nav is unavoidable) |
| Internal-host policy bypassed (authed SSRF) | high | `block_internal_hosts` threaded into **both** scope checks (fetch URL + navigated origin); producer also refuses under the policy |
| POST under "read-intent only" label | med | claim softened — transport can't verify; read-intent is the producer's responsibility (assertReadOnlyPath + operator) |
| Auth cookies inherited by browser child procs via env | med | `delete BOB_BROWSER_DRIVER_INIT` before Chromium launches → renderers don't inherit cookies/proxy creds. Driver's own `/proc/environ` is the documented same-UID boundary (#131) |
| Plan doc names the client + ships in the package | P1 | genericized — no client/engagement specifics in `docs/OFFENSIVE_MASSREAD_PRODUCER_PLAN.md` |

+5 source-contract tests for the above. The real-Chromium end-to-end differential still passes with redirect-manual + streaming + commit-nav. Full `npm test` green (3273 pass, 0 fail; 2 pre-existing skips).

@coderabbitai review
@codex review

vmihalis:
## PR1 round-3 — round-2 dug a deeper, legitimate layer; all addressed

Round-2 didn't just re-state round-1 — the bots found a real second layer in the trusted transport (exactly why this is its own isolated PR). Every finding was verified against the code before fixing. Disposition:

| Round-2 finding | Sev | Disposition (`<commit>`) |
|---|---|---|
| **DNS-rebinding TOCTOU** — Node validates the IP, Chrome re-resolves independently → public IP to the check, internal IP to the browser | P1 / high (×3) | **Fixed** — target host resolved once in Node; Chrome launched with `--host-resolver-rules=MAP <host> <validated-ip>` so it connects to the checked IP. Under `block_internal_hosts`, an unpinned host is refused fail-closed. Skipped under a proxy (DNS resolves there). |
| **Auth cookies shipped through the child process env** | high (×2) | **Fixed** — cookies no longer touch the env at all. They're sent after ready over stdin via a new server-side-only `set_auth_cookies` command; the env scrub remains for proxy creds. |
| **Cookie injection is context-level / cookies not scope-checked** | high | **Fixed** — `set_auth_cookies` validates EACH cookie's target host against `target_domain` (via `assertSafeRequestUrl`) before it reaches the context; one out-of-scope cookie fails the whole session closed. |
| **Body cap counts UTF-16 code units, not bytes** ("2 MiB" ≈ up to 4 MiB) | med (×2) | **Fixed** — cap measured on `Uint8Array.byteLength`/`subarray`, enforced while streaming; "2 MiB" now means 2 MiB. |
| **Timeout reports to caller but doesn't abort the in-page fetch** | med | **Fixed** — in-page `AbortSignal.timeout(timeout + grace)` so the renderer request is cancelled; the grace lets the deterministic wall-clock timeout still win the race. |
| **Page-controlled fetch** (page could override `window.fetch`) | P1 | **Accepted residual, justified.** A native-fetch capture was implemented and **reverted** — it's unreliable under the Patchright stealth driver (its `fetch` wrapper rejects a detached/bound call → "Failed to fetch") and broke the transport. Carrying the real Chrome TLS fingerprint *requires* the page network stack (the gap this transport closes). Compensating controls: `waitUntil:"commit"` minimizes page-script execution before the fetch, and the authed-vs-control **differential** bounds integrity (page code can't tell which arm it serves; a target forging a vuln against itself isn't a coherent threat). A single capture is evidence, not proof. |
| **"Server-side-only is convention, not enforced"** | med | Enforced by test: no `bob_browser_*` tool wrapper references `authed_fetch` *or* `set_auth_cookies` (and the agent `evaluate` sandbox stays fully closed). |
| **CodeRabbit: security contracts need behavioral coverage, not just source regex** | Major | **Addressed** — added 5 real-Chromium e2e behavioral tests: authed-vs-control differential, `redirect:manual` → status 0, oversized-body byte-cap + truncation flag, out-of-scope cookie rejection, out-of-scope URL refusal. |

Full `npm test` green. The focused file passes 26/26 including all 5 real-Chromium e2e tests.

@coderabbitai review
@codex review

vmihalis:
## PR1 round-4 — round-3 found real bugs in the round-3 DNS-pin; all fixed

Round-3 added a DNS-pin, and the review correctly found that my pin was incomplete. Every finding verified against the code, then fixed:

| Round-3 finding | Sev | Disposition |
|---|---|---|
| **Pin records a launch-time IP without internal-validation; `authed_fetch` presence-only check (`pinnedHosts.has`) lets a private pinned IP satisfy `block_internal_hosts`** | high (×2) | **Fixed** — the pin now records `{ address, internal: isBlockedInternalHost(address) }`. Under `block_internal_hosts`, `authed_fetch` refuses an **unpinned** host *and* a host whose **pinned IP is internal**. This closes the rebind where a host resolves public at check-time but the pinned (connected) IP is private. |
| **DNS rebinding not closed even under `block_internal_hosts`** | P1 / high | **Addressed** — `--host-resolver-rules` makes Chrome connect to the launch-resolved IP (no independent re-resolution); combined with the pin-internal guard above, the TOCTOU is closed for the pinned host. |
| **Under a proxy, pinning is disabled → `block_internal_hosts` becomes advisory** | med | **Fixed** — `authed_fetch` now **refuses** when proxied *and* `block_internal_hosts` (DNS resolves at the proxy; we can't pin/vet it, so we fail closed instead of silently degrading). |
| **Cookie scope validated on a derived URL but the original cookie object is written** | med | **Fixed** — reject the ambiguous both-`url`-and-`domain` form; validate the exact effective host (`url.host`, or the de-dotted `domain`), so the validated target == what is written. |
| **Timeout reports to the caller but doesn't abort the authenticated fetch** | med | **Fixed** — `AbortSignal.timeout` now fires **at** the deadline (was deadline+grace), cancelling the in-flight request; the in-page catch converts the abort to a deterministic `{ __timeout: true }` sentinel (no fragile message matching). |
| **Design doc ships an engagement-specific name** | P1 | **Fixed** — removed. |
| "Server-side-only is convention, not enforced" | med | **Disposition** — enforced by test: no `bob_browser_*` tool maps to `authed_fetch` *or* `set_auth_cookies`, and the agent `evaluate` sandbox stays fully closed. These are driver-internal commands, not registered capabilities. |
| "Trusted/server-side but runs in the page renderer" | med | **Disposition** — same as the documented page-controlled-fetch residual (real TLS fingerprint requires the page stack; a native-fetch capture breaks under Patchright). Compensating controls: commit-nav + the authed-vs-control differential. |

New behavioral e2e: a loopback-resolving target under `block_internal_hosts` is refused. Full `npm test` green; focused file **28/28** including all real-Chromium e2e.

@coderabbitai review
@codex review

vmihalis:
## PR1 round-5 — the two genuinely-new findings fixed; the rest are re-states of logged dispositions

Round-4's review had two findings that were *new and correct*; both fixed. The remainder are the advisory bots re-anchoring items already fixed or dispositioned in earlier rounds.

### Fixed (new, correct)

| Finding | Sev | Fix |
|---|---|---|
| **The DNS pin is opt-in — the default path (no `block_internal_hosts`) is a credentialed fetch gated only by hostname, so a rebound in-scope subdomain rides to any IP** | high (persistent ×3) | **Fixed** — the pin is now **required on every `authed_fetch`** (not just under `block_internal_hosts`). On the non-proxied path an unpinned host is refused unconditionally, so a credentialed request can never ride a rebound hostname. Producer contract: start the session with `target_url` on the host you'll read (that's the pinned one). New e2e proves an in-scope-but-unpinned host is refused on the default path. |
| **My disposition claim "page cannot tell which arm it serves" is false — the *server* sees the auth cookie** | med (new) | **Fixed reasoning** — the differential is NOT tamper-proof; the target can distinguish arms and could serve forged data. The honest argument: a target forging a vuln against *itself* gains nothing, the per-tool ceiling bounds a forged result, and a single capture is evidence, not proof (operator corroborates for integrity-sensitive use). Comment + doc corrected. |

### Re-states of already-fixed / dispositioned items

- **"Timeout doesn't abort the fetch"** — fixed in round-4: `AbortSignal.timeout` fires at the deadline and the in-page catch returns a `{__timeout:true}` sentinel. The fetch *is* aborted; the wall-clock race is a backstop for a fully-hung `page.evaluate`.
- **"Cookie validated on a derived URL"** — fixed in round-4: the ambiguous both-`url`-and-`domain` form is rejected and the exact effective host is validated, so validated-host == sent-to host.
- **"Server-side-only is convention"** — dispositioned: enforced by the IPC topology (only the in-process producer can write the driver's stdin) **and** by test (no `bob_browser_*` tool maps `authed_fetch`/`set_auth_cookies`; the agent `evaluate` sandbox stays closed). Not a registered agent capability.
- **"Runs in the page renderer"** — the documented page-controlled-fetch residual (real TLS fingerprint requires the page stack; a native-fetch capture breaks under Patchright). Compensating control: commit-nav; integrity bounded as above.

Full `npm test` green; focused file **29/29** including all real-Chromium e2e (differential, redirect-not-followed, byte-cap truncation, cookie-scope rejection, **unpinned-host refused on the default path**, **loopback refused under `block_internal_hosts`**, out-of-scope refusal).

@coderabbitai review
@codex review

vmihalis:
## PR1 round-6 — one new finding fixed (auth-log leak); the priming-redirect is dispositioned (a fix would defeat the transport)

Round-5's review had two new findings. One is a clean fix; the other I initially "fixed" with a network-route interceptor and then **reverted** — because the repo already has a tested invariant forbidding exactly that, for a good reason.

| Round-5 finding | Sev | Disposition |
|---|---|---|
| **Credentialed `authed_fetch` request (cookies/headers) captured into the agent-readable network log / record buffer** | med | **Fixed** — a per-op flag makes `attachListeners()` skip recording the priming nav + the fetch, so the producer's auth material never enters `this.requests` / `recordedRequests` (which is flushed to the audit trail agents read). New e2e proves the request is absent from the network log. |
| **Priming `page.goto` follows redirects / the primed page can initiate loads → a 3xx to an internal host fires one request before the drift guard** | P1 | **Dispositioned (NOT fixed with an interceptor).** The only way to block page-initiated loads is a catch-all `context.route` interceptor — which `browser-driver-tools.test.js` explicitly forbids because it aborts page-decided subresource loads (Kasada/Akamai anti-bot, CDN bundles, OAuth callbacks), i.e. the exact WAF-protected flows this transport exists to survive. Fixing this would **defeat the transport's purpose**. Bounded instead by: the producer **refuses** under `block_internal_hosts` (so this is a direct-driver defense-in-depth path), the pin connects the target host to a validated IP, the origin-drift guard refuses the fetch after any redirect, and the scope check validates the URL's resolved IP under `block_internal_hosts`. (I tried the interceptor; CI's own invariant test caught it — good.) |
| **Single-IP pin removes DNS fallback** | P2 | **Disposition** — intentional: falling back to other resolved IPs reopens the rebind window the pin closes. Sessions are short-lived. |

Everything else is advisory bots re-anchoring items fixed/dispositioned earlier (require-pinned, timeout-abort, cookie validation, differential prose, server-side-convention, page-renderer) — logged in prior comments.

Full `npm test` green; focused file **32/32** including 8 real-Chromium e2e behaviors (differential, redirect-not-followed, byte-cap truncation, cookie-scope rejection, unpinned-host refused on default path, loopback refused under `block_internal_hosts`, **auth request absent from the network log**, out-of-scope refusal). The no-`context.route` invariant test passes.

@coderabbitai review
@codex review

vmihalis:
## PR1 round-7 — enforce cookie-only auth + session-wide log suppression (the last fixable items)

Round-6's review found two more real items rooted in the same truth — *a credentialed page-world fetch inherently exposes header-auth and a credentialed context retains cookies*. Rather than whack-a-mole, round-7 **enforces the safe path**:

| Round-6 finding | Sev | Fix |
|---|---|---|
| **`headers` param is built in the page world → a producer's `Authorization`/`Cookie` header could be read by a page-overridden fetch** | P1 | **Fixed (enforced)** — `authed_fetch` now **rejects** credential headers (`Authorization` / `Cookie` / `Proxy-Authorization`). Auth must flow via `set_auth_cookies`, which the browser attaches at the network layer — invisible to page JS. So a page-overridden fetch sees a request but no secret. |
| **Auth-log suppression cleared post-op while the credentialed context persists → cookies ride (and get logged on) later page requests** | P1 / med | **Fixed** — once `set_auth_cookies` injects cookies, the WHOLE session is marked credentialed and request logging is suppressed for its entire life (not just the op). A credentialed session is producer-only; agent recon runs in uncredentialed sessions. |
| **Page-controlled residual is more durable via service workers** | med | **Disposition (extended)** — a target could also use a service worker to intercept. Same bound as the page-controlled-fetch residual (incoherent self-sabotage + ceiling + evidence-not-proof). Crucially, credentials are now NOT exposed by this residual: cookie auth is browser-attached and credential headers are rejected, so a page/SW override sees a request but no bearer/cookie. |

Re-states of already fixed/dispositioned items (require-pinned, pin-internal, timeout-abort, cookie-scope, differential prose, **priming-redirect** — which can't be fixed without the forbidden `context.route` interceptor, server-side-convention, DNS-fallback) are logged in prior comments.

Full `npm test` green; focused file **35/35** including 10 real-Chromium e2e behaviors (now incl. **auth request absent from the log**, **Authorization header rejected**).

This addresses every *fixable* finding; the remainder are dispositioned residuals (page-controlled-fetch incl. SW, priming-redirect, server-side-convention, DNS-fallback), several backed by existing tested invariants. Intending to resolve threads and merge after this pass unless a genuinely-new fixable issue appears.

@coderabbitai review
@codex review

vmihalis:
## PR1 final disposition — merging (CodeRabbit clean + required CI green; advisory threads resolved with justification)

This trusted, WAF-piercing, credentialed browser transport was isolated into its own PR specifically so it could be reviewed hard. It was — across 7 rounds. **CodeRabbit (the authoritative review App) has 0 unresolved threads and passes; required checks `test (20)`/`test (22)` pass.** The remaining unresolved threads are all from the advisory bots (Brutalist CI + Codex connector) re-anchoring items already **fixed** or **dispositioned with justification**. Per the repo merge rule, merging over advisory findings is allowed with a logged rationale — here it is.

### Fixed across rounds (verified, with tests)
- **Redirect-follow → off-scope auth leak** → `redirect:"manual"` (3xx surfaces as status 0; oracle fails closed).
- **Unbounded body buffering / UTF-16 "byte" cap** → stream-read with a true **byte** cap (`Uint8Array.byteLength`).
- **Auth cookies in the process env** → moved to a stdin `set_auth_cookies` command; secrets never touch the env.
- **Cookies not scope-checked / ambiguous form** → each cookie's effective host validated vs `target_domain`; `url`+`domain` both-set rejected.
- **DNS rebinding** → Chrome is DNS-pinned (`--host-resolver-rules`) to the Node-validated IP; the pin records internal-ness; an **unpinned host is refused on every call** (not opt-in) and a **pinned-internal IP is refused** under `block_internal_hosts`; proxied + `block_internal_hosts` is refused (can't pin through a proxy).
- **Hung fetch not aborted** → in-page `AbortSignal.timeout` at the deadline + deterministic `{__timeout}` sentinel.
- **Auth request captured into the agent-readable log / audit trail** → the trusted op is not logged, and once cookies are injected the **whole credentialed session** is suppressed.
- **Header-auth exposed to the page world** → `authed_fetch` **rejects** `Authorization`/`Cookie`/`Proxy-Authorization`; auth flows only via browser-attached cookies (invisible to page JS).
- **False reasoning** ("page can't tell which arm") → corrected.
- Plus: env scrub before launch, origin-drift guard, `waitUntil:"commit"`, agent `evaluate` sandbox unchanged, no MCP tool maps `authed_fetch`/`set_auth_cookies`.

### Dispositioned (advisory bots will keep re-flagging these; each is sound)
- **Priming-redirect under `block_internal_hosts`** (P1): the only block is a catch-all `context.route` interceptor, which `browser-driver-tools.test.js` **forbids** (it aborts Kasada/Akamai/CDN/OAuth subresource loads — the WAF-protected flows this transport exists to survive). Fixing it would defeat the transport. Bounded by: producer refuses under `block_internal_hosts`, the pin, the origin-drift guard, the scope check. (I tried the interceptor; that invariant test caught it; reverted.)
- **Page-controlled fetch incl. service workers**: inherent to using the real Chrome stack (a Node-side `context.request` is CF-403'd — the gap this transport closes); a native-fetch capture breaks under Patchright. Credentials are NOT exposed (cookie-attached + headers rejected); integrity is bounded (a target forging a vuln against itself is incoherent; per-tool ceiling; evidence-not-proof).
- **"Server-side-only is convention"**: enforced by the IPC topology (only the in-process producer can write the driver's stdin) and by test (no `bob_browser_*` tool maps these commands).
- **DNS-fallback removed by single-IP pin**: intentional — fallback reopens the rebind window; sessions are short-lived.
- **Broad (leading-dot) domain cookie**: the de-dotted apex is validated against scope; under the standard domain+subdomain scope model, apex-in-scope ⟹ subtree-in-scope, and the producer supplies its own (non-malicious) session cookie.
- **Record/audit disabled for credentialed sessions**: intentional and documented — that's the point (don't record auth traffic); credentialed sessions are producer-only.
- **`block_internal_hosts` defaults false / per-fetch latency**: the policy is the operator's deliberate choice (pin is always required regardless); the goto is the same-origin requirement and the delay is anti-detection; mass-read is low-volume.

Resolving the advisory threads and merging. Follow-up PR2 (the `bob_http_massread_confirm` producer) builds on this transport.

PR #156, opened 2026-06-23
feat(offensive): bob_http_massread_confirm — broken-auth/BFLA mass-read producer (PR2)

vmihalis:
## What & why

Adds the **first signed producer for the broken-auth / BFLA mass-read class** — an under-authorized identity (leaked/guessable credential, low-priv account) bulk-reads a sensitive collection it should not. This class had **no producer** on the signed offensive rail; a client pentest surfaced exactly this (a listing endpoint returning many production PII records to a caller that should see only its own). Builds on **#155**'s `authed_fetch` WAF transport. Spec: `docs/OFFENSIVE_MASSREAD_PRODUCER_PLAN.md`.

## The oracle — `safe_oracle.kind: "differential_response"`, ceiling **HIGH**

Two requests, both **through the WAF via the real-Chrome transport**:
- **attacker** (the under-authorized identity's cookies, read server-side from a named auth profile) → must be `2xx` + `record_count ≥ 2` + ≥1 sensitive field present.
- **control** (fresh, uncredentialed) → must be **denied**: `401/403`, or a `2xx` with 0 records / no sensitive shapes.

A control that *also* reads the bulk data ⇒ a **public endpoint, not an authz break** ⇒ no mint (`blocked_by_design`). WAF/rate-limit on the attacker arm ⇒ fail closed (a block page is not "0 records").

## Witness — dual output (masked rail + opt-in raw)

1. **Signed row (`offensive-runs.jsonl`) — ALWAYS masked.** Only `record_count` + sensitive **field-name buckets** (email/phone/financial/government_id/dob/postal_address/other_sensitive — never the target's literal column names) + value-**shape booleans** + differential statuses. Screened by `sensitiveShapesPresent`, **fail-closed** (`blocked_operator_pii` if any value shape sneaks in). The raw bodies are derived-then-discarded. **Non-negotiable: no raw PII on the signed rail** (the same tool runs against third-party bug-bounty targets where harvesting strangers' data is forbidden).
2. **Full raw capture — opt-in, operator-only.** Written ONLY under the env gate `BOB_MASSREAD_OWNER_AUTHORIZED=1`, to `sessionDir/massread-evidence/<run_id>.json` (mode 0600, outside the repo and the signed rail, operator deletes after the engagement). **Not an agent argument** — the agent can never enable PII capture (it's in the forbidden-input set).

## Safety invariants (the review checklist)

- `demonstrated_severity` stamped from the frozen registry inside `buildAndSignOffensiveRow` — never agent-supplied.
- Producer **refuses under `block_internal_hosts`** (mirrors `xss_confirm`; the browser can't enforce the session SSRF policy). Every request via `auditConfirmRequest`.
- Auth cookies read **server-side** from the named profile → `set_auth_cookies` (stdin, never env, page-invisible); each cookie scope-validated by the driver.
- **Minimal agent surface:** the tool accepts only `target_domain` + `surface_id` + an optional `auth_profile` *selector* (a registry handle, not credentials). No URL / endpoint / cookie / token / header / body / record / severity / `owner_authorized`. The listing endpoint is server-derived from the routed surface and re-validated (scope + read-only path) before any read.

## v1 scope (honest)

**GET + cookie-expressible identity.** POST-body listing and **bearer-token** identities are deferred to v2 (both return `blocked_by_design`) — the latter because #155's `authed_fetch` is cookie-only by design (it rejects `Authorization` in the page-world headers param). v1 still covers a real, common class. The motivating `POST`+`Bearer` shape needs v2 (POST-body value-blind resolution + a secure bearer-injection transport primitive).

## Integrity boundary (honest, not closed here)

The row MAC is tamper-evident against an MCP/guarded-Bash-confined agent but **not** un-forgeable by a same-UID actor (the #131 boundary, bounded to a fabricated HIGH by the frozen ceiling). The authed-vs-control differential is target-distinguishable (the target can tell the credentialed arm) — a single capture is **evidence, not proof**; the operator corroborates for integrity-sensitive use. Inherits #155's accepted page-controlled-fetch residual (real TLS fingerprint requires the page network stack; mitigated by `waitUntil:"commit"` + the differential).

## Tests / gates

New `test/offensive-massread-producer.test.js` — **20/20** (positive mint; negatives: control-not-denied, count<min, no-sensitive-shape, `block_internal_hosts` refused; masked-witness round-trip proving no raw PII on the signed row; `demonstrated_severity:"high"`; `owner_authorized` gating). Seeded-driver (mock transport) — a live browser/target smoke is the operator's next step. **All gates green:** `check:syntax` · `test:mcp` 3017/0 · `test:prompts` 202/0 · `test:install` 16/0 (real server = exactly 160 tools) · authority-inventory no-drift · lean pack 3.34 MB (under the 3.4 MB gate).

## Release Notes
* **New Features**
  * Added `bob_http_massread_confirm` differential mass-read confirmation with attacker-vs-control semantics, masked signed-row evidence, and an operator opt-in gate for raw captures.
* **Security**
  * Session read guards now block direct access to `massread-evidence`.
  * Updated demonstrated-severity ceiling rules for the new tool.
* **Documentation**
  * Updated evaluator prompts/skills with `differential_response` evidence-oracle mapping and minting/recording guidance.
* **Tooling**
  * Extended tool allow-lists, authority inventory, and tool-budget checks for enablement.
* **Tests**
  * Added/expanded coverage and updated registry/manifests/tool-count assertions.


vmihalis:
@coderabbitai review

vmihalis:
@codex review

vmihalis:
## Re-review request — soundness hardening pushed (`2a47df2` → `d0b9fe9`)

Thanks for the first pass. The findings all landed on real ways this HIGH-stamping producer could sign a **false HIGH** or mishandle the raw-PII path. Applied as `d0b9fe9` — every fix tightens a gate or hardens the capture; none weakens the masking rail. Highlights:

**False-HIGH gaps closed**
- **Truncation → fail closed** on *both* arms — a truncated control mis-scored as "0 records → denied" (false HIGH); a truncated attacker → false negative. Unparseable attacker 2xx body also fails closed.
- **Per-arm cache-buster** (`_cb` nonce) so a shared cache can't cross-serve one arm's body to the other and forge the differential; `arm_order` recorded in the hash-bound bundle.
- **MINT GATE now requires a real PII *value* shape** (`pii_shape_present`), not a substring field-*name* match — `emailVerified:true` no longer lifts a row. Field-name matcher tightened to camelCase/word-boundary tokens (`themailbox` ≠ `email`).
- **Control fail-closed on ambiguity** — a WAF/5xx/404/unparseable control is `control_inconclusive`, never scored "denied" and never "also reads."

**Honest framing (the deepest finding)**
- The authed-vs-**unauthenticated** differential proves "credential X bulk-reads PII an unauth client is denied" — *not* that X is under-privileged. Signed booleans are now `credentialed_bulk_pii_read` / `unauthenticated_control_denied` / `pii_value_shapes_present` (no false `bfla_proven`), with a loud **operator contract** in the tool description + evaluator prose (`auth_profile` = the *leaked/under-privileged* credential; a fully-authorized one is misuse → false positive). True cross-tenant BFLA (a second authenticated *victim* arm) is the documented v2 oracle.

**Raw-PII path hardened**
- `writeFullCapture` now uses the same symlink discipline as the signed rail: `realpath` the session root, `lstat` refuse pre-existing symlinks, post-mkdir realpath-equality, `O_CREAT|O_EXCL|O_NOFOLLOW` mode 0600.
- `massread-evidence/` added to the session read-guard (Claude + Kimi) — the agent can never `Read` the raw capture; only the operator does.

**Cleanup:** runArm null-guard, recorded-query preserved on the fetch URL (canonical signed target stays query-stripped), cookie-split bound, dead `rejectInvalidArguments` removed.

Gates (local, full): `node --test` **31/0** (+11 new) · `test:mcp` **3028/0** · `test:install` **16/0** (160 tools) · `test:prompts` **202/0** · read-guard **116/116** · package gate **20/0** (CI-equivalent tree).

@codex review
@coderabbitai review

vmihalis:
## Round-2 soundness pushed (`d0b9fe9` → `f922a81`) — disposition of all open threads

The first-round review left **24 unresolved threads**. Re-reading them, several were **not** closed by `d0b9fe9` — genuine false-HIGH / false-negative paths. `f922a81` fixes them. Full ledger:

### Fixed in `f922a81`
- **PII value detached from the counted records → false HIGH** (`:305` brutalist, `:305` Codex P1, `:592`, `:223`). The mint counted the largest array and scanned the **whole body** for a PII value, so `{ data:["a","b"], support_email:"x@y" }` + a 401 control minted HIGH with no PII in any record. Now: `extractRecords` counts **arrays of objects only** (a scalar array can't inflate the count or host the value); `deriveMaskedSummary` detects PII **per-record over record values only**; the mint gate **re-requires ≥1 sensitive field-name bucket**. All three signals now bind to the **same** counted records. *(test: metadata-email no-mint, scalar-array→0 records, object-array pick, field-name requirement.)*
- **Narrow PII vocab → SSN/card mass-reads never mint** (`:154`, `:156`). Replaced the email/phone/iban regex with the repo's own `detectPiiShapes` (email/phone/**SSN**/**Luhn card**) ∪ IBAN, run per-record (also resolves the 256 KB whole-body truncation — each record is scanned independently, up to 1000 records). *(test: SSN mints, credit-card mints.)*
- **Cookie reconstruction could forge an identity** (`:191`). A stored value `abc; role=admin` re-split into a forged `role=admin` cookie. `auth.js` now persists a **structured `cookie_jar`** (Bob-local metadata); the producer issues cookies verbatim → an embedded `;` stays one cookie. *(test: forge prevented + legacy fallback.)*
- **Control 403 from a WAF challenge counted as denial** (`:606` Codex P1). A 403 is a clean unauthenticated denial only with a **parseable JSON** body; an HTML/unparseable block-page 403 → `control_inconclusive` (fail closed). *(test: 403 HTML inconclusive vs 403 JSON denied.)*

### Already fixed in `d0b9fe9` (these threads predate it)
- `:389` truncation ignored → both arms fail closed. `:568` WAF on attacker arm only → control WAF check added. `:482` recorded query stripped → preserved on the fetch URL (signed target stays query-stripped).

### Resolved with justification (not code changes)
- **"authed-vs-unauthenticated ≠ under-privileged"** (`:551`, `:592` design, `bob-http-massread-confirm.js:10`). Correct observation. v1's resolution is the **honest framing**: the signed booleans assert only `credentialed_bulk_pii_read` / `unauthenticated_control_denied` / `pii_value_shapes_present` — **never** `bfla_proven`/`under_privileged` — under a loud **operator contract** (`auth_profile` = the leaked/under-privileged credential; a fully-authorized one is misuse). The HIGH reflects the **bulk-read impact**, and the grader + 3-round verification corroborate before anything is reported. True cross-tenant proof (a second authenticated *victim* arm) is the documented **v2** oracle. CodeRabbit re-reviewed this framing on `d0b9fe9` and approved it.
- **Codex adapter can read the raw capture** (`:681` Codex P1). True, and pre-existing: the Codex adapter has **no read-guard hook** for *any* session artifact. The opt-in raw capture's **primary** control is the operator env gate (`BOB_MASSREAD_OWNER_AUTHORIZED`, agent-unsettable, enabled only for client-owned data the operator is already authorized to hold); the Claude/Kimi read-guard is defense-in-depth. Documented in-code at `writeFullCapture`. The **always-masked signed rail never carries raw PII on any adapter** — that guarantee is unconditional.

Gates (local, full): `node --test` **40/0** (+9) · `test:mcp` **3037/0** · `test:prompts` **202/0** · read-guard **116/116** · package gate **20/0** (CI-equivalent tree).

@codex review
@coderabbitai review

vmihalis:
## Round-3 (`f922a81` → `bd68d7e`) — disposition of the 15 new threads

Thorough pass — thank you. Most were genuine. Disposition:

### Fixed in `bd68d7e`
- **PII not bound to the sensitive field / boilerplate / single-subject false HIGH** (`:354`, `:658`, `:657`, `:305`, `:268`). Reworked the oracle: a PII value counts **only** when it sits in a sensitively-**NAMED** field of a counted record, and only **DISTINCT** `(bucket,value)` pairs count. Gate is now `pii_bearing_count ≥ 2 && distinct_pii_count ≥ 2`. A stray metadata email, a field merely *named* like PII, a **constant** boilerplate field (same support email per row), and a **single self-record** (values repeated through a nested array) all fall below the floor. *(tests: constant-field, single-subject, nested self-record → no mint.)*
- **Control read the bulk body but scored "denied"** (`:677`, `:669`). Restructured: a control whose body parses to ≥2 records **read** the collection regardless of HTTP status (a 401/403 that still returns the bulk body was **not** denied) → public, no mint. *(tests: control 401/2xx-with-bulk-body → `control_also_reads_bulk`.)*
- **JSON-403 WAF mints false HIGH** (`:675`). A 403 is ambiguous (WAF/bot-challenge vs authz) and is **never** scored denied — only an explicit 401 (or a 2xx returning <2 records) is. *(test: JSON-403 → `control_inconclusive`.)*
- **`writeFullCapture` symlink anchor** (`:460`). Now anchors to `realpath(sessionsRoot()) + assertSafeDomain(domain)` exactly like `offensive-capture-writer`'s `resolveCaptureDirSecure` — a symlinked **session dir** can no longer redirect the raw-PII write. (The signed row already anchored this way; only the raw capture didn't.)
- **Fetched URL vs signed target divergence** (`:553`, `:498`). Dropped the recorded-query resurrection: v1 fetches the resolved endpoint as-is, so the fetched URL and the signed `canonicalTarget` are the **same** origin+path. Query-routed search surfaces are v2.

### Resolved with justification
- **`auth_profile` is unvalidated → any account can mint a HIGH** (`:517` 🔴, `:713` 🔴). This is the irreducible "authed-vs-anon ≠ under-privileged" point — correct, and it only fully closes with v2's victim-arm (a second authenticated victim proving cross-tenant access). **The operator made an informed call to keep the HIGH ceiling**: this runs as a fully-autonomous pipeline, so a MEDIUM ceiling would *permanently* under-report genuine leaked-credential mass-reads (the motivating client case: a public token → 51 real records) with no human to escalate. The "is this credential under-privileged" judgment is carried by the **evaluator agent** (it knows the credential's provenance — it found/stored it) under the loud operator contract, plus the context-aware grader; the signed booleans never claim `bfla_proven`/`under_privileged`. The round-3 fixes above narrow the HIGH to genuine ≥2-distinct-subject bulk reads with a denied control, shrinking the false-HIGH surface to that one irreducible authz-semantics gap. **v2 victim-arm is the tracked follow-up** to make it machine-proven.
- **Cookie cross-host scope** (`:208`). v1 binds the stored jar's cookies to the routed endpoint origin; the driver scope-validates every cookie against `target_domain`. Per-subdomain scope fidelity (a `www`-scoped cookie not sent to `api`) needs the jar to carry each cookie's captured domain/path — deferred to v2 alongside the victim-arm. Documented limitation.

Gates (local, full): `node --test` **46/0** (+11) · `test:mcp` **3043/0** · `test:prompts` **202/0** · read-guard **116/116** · package **20/0** (CI-equivalent tree).

@codex review
@coderabbitai review

vmihalis:
## Round-4 (`bd68d7e` → `c21d3a5`) — disposition of the 9 threads

Real finds — thank you. One was a genuine false-HIGH the round-3 fix didn't actually close.

### Fixed in `c21d3a5`
- **🟠 P1 — same-subject multi-field defeats the distinct-subject gate** (`:387`). Correct, and the important one. `distinct_pii_count` counted DISTINCT `(bucket,value)` pairs **across** buckets, so one subject carrying email+phone (or email+ssn) repeated across rows reached 2 from `(email,v)+(ssn,v)` and minted — the exact self-record case the gate was meant to suppress. Reworked: `distinct_pii_count` is now the **MAX distinct values in any SINGLE identity bucket** (≥2 distinct emails, or phones, …) — a robust lower bound on distinct subjects. One subject's several PII fields stay at cardinality 1 → below the floor. *(test: email+ssn one-subject → `distinct_pii_count:1`, no-mint, with BOTH buckets populated so it's a true multi-bucket case; + a two-distinct-emails → mints regression.)*
- **🟡 Empty 2xx control scored inconclusive** (`:706`). A 204 / empty-200 unauth control read **no** bulk collection — a clean denial — but `parse_ok` was required, so it fell to `control_inconclusive` and missed the leak. Now an empty 2xx denies; an unparseable **non-empty** 2xx (HTML shell / WAF page) **stays** inconclusive (parseability is still not an auth oracle). *(tests: empty 200 + 204 → mint; HTML-shell 2xx → inconclusive.)*
- **🟡 `writeFullCapture` TOCTOU parent-dir symlink swap** (`:522`). Replaced the post-mkdir realpath re-check (realpath **follows** a symlink and passes) with an **`O_NOFOLLOW` directory-fd inode pin**: `O_DIRECTORY|O_NOFOLLOW` fails closed (ELOOP) if `massread-evidence` was swapped to a symlink after the lstat, and `fstat` dev+ino binds the fd to the trusted inode. **Honest residual:** core `fs` has no `openat(2)`, so the leaf is still opened by pathname — the sub-µs same-UID re-swap window is the conceded **#131** boundary (a same-UID actor already reads the 0600 signing key) and additionally needs the operator capture gate **enabled**. Bounded, not open — consistent with your own medium rating.
- **🟡 Broken `bob_auth_store` instruction** (`evaluator-agent.md:49`). Right — `bob_auth_store` is in the `auth` bundle, not `evaluator-web`, so the per-surface evaluator can't call it. Dropped the "store cookies via `bob_auth_store`" line; a missing profile now routes to a `blocked_prereqs[]` `auth_missing` entry (the existing prereq pattern). Regenerated the agent/adapter surfaces.
- **Bonus:** removed a stray **NUL byte** committed in `bd68d7e` (a round-3 heredoc artifact on the old `distinctValues.add` line — harmless as a key separator, but it made the file scan as binary).

### Deferred — precision-safe recall gaps (they make the producer MISS leaks, never over-mint; v1 boundaries)
- **DOB/address value-shapes** (`:383`). The value-shape gate is deliberately high-precision (email/phone/ssn/card); DOB/postal-address regexes are noisy and need name+value corroboration to avoid date/number false positives. A DOB/address-only leak under-reports — the safe direction — until v2 adds corroborated shapes.
- **Count attacker bodies on error status** (`:685`). v1 mints only on a clean 2xx attacker read; a 403-with-leaked-body is a real but separately-classed issue. Requiring 2xx is the conservative bound (fewer mints), so this is a recall deferral, not a soundness gap.
- **Preserve redirect-to-login controls** (`:474`). A status-0/`opaqueredirect` control is genuinely ambiguous (denied-redirect vs network error); v1 treats it as inconclusive (precision-safe). v2 can parse the redirect `Location` for `/login`.

### Resolved with justification — the design ceiling (operator-decided)
- **`auth_profile` unvalidated → authed-vs-anon mints HIGH** (`claims.js:797` 🟠, `:705` 🟠). The same irreducible "authed-vs-anon ≠ under-privileged" point from round-3. **The operator made an informed call to keep the HIGH ceiling** for a fully-autonomous pipeline (a MEDIUM ceiling would permanently under-report genuine leaked-credential mass-reads — the motivating client case: a public token → 51 real records — with no human to escalate). The evaluator (which knows the credential's provenance) carries the under-privileged judgment under the loud operator contract; the signed booleans never claim `bfla_proven`/`under_privileged`. Round-4 narrows the HIGH to genuine ≥2-distinct-**subject** bulk reads with a denied control. **v2's victim-arm** (a second authenticated victim denied while the attacker reads its data) is the tracked follow-up that makes it machine-proven.

Gates (local, full): `node --test` **50/0** (+4) · `test:mcp` **3047/0** · `test:prompts` **202/0** · `check:syntax` clean.

@codex review
@coderabbitai review

vmihalis:
## Round-5 (`c21d3a5` → `dd3f377`) — disposition of the 9 threads

You were right — the round-4 per-bucket counter was still defeatable three ways. Round-5 redesigns it properly.

### Fixed in `dd3f377`
The distinct-subject count is now the **MAX, over every (concrete FIELD KEY, PII SHAPE) pair, of the distinct NORMALIZED values for that pair across records**. That single change closes all three residual false-HIGHs:
- **Intra-bucket multi-field** (`:393` P1, `:405` 🟡). `email` and `recovery_email` both map to the `email` *bucket* but are distinct *fields* — keyed per concrete field, each is cardinality 1, so one subject's two same-bucket fields no longer inflate. *(test: email+recovery_email one subject → no-mint.)*
- **Unnormalized values** (`:394` 🟡). Values are canonicalized before counting — email lowercased, phone/ssn/card digits-only, IBAN upper+stripped — so `Alice@X.com`/`alice@x.com` and `(555) 123-4567`/`555-123-4567` collapse to one subject. *(test: case/format variants → no-mint.)*
- **Whole-JSON structured value** (`:394` P1). Only the **extracted PII token** is keyed, not the serialized field — a `{ value: "a@x.com", label: "billing"|"shipping" }` with a varying sibling stays at cardinality 1. *(test: structured value → no-mint.)*
- **(bonus, avoided a new defeat)** per **shape** within a field too, so a single field holding one subject's email AND phone is two shapes of one subject, never summed to 2. *(test: two shapes one field → no-mint.)*
- **🟡 status-0 audit crash** (`:448`). `authed_fetch` surfaces a manual redirect (unauth redirect-to-login control) as status 0, which `Number.isInteger` let through to the http-audit normalizer → `probe_audit_failed` **crashed the run** instead of failing closed. Now any status outside [100,599] coerces to null → `control_inconclusive`. *(test: status-0 control → inconclusive, no throw.)*
- **🔴 read-guard symlink-alias bypass** (`session-read-guard.sh:101`). Correct — `BLOCKED_DIRS` was checked against the literal `resolved.parts`, so `/tmp/e -> <session>/massread-evidence` let `Read /tmp/e/<run>.json` slip past the raw-PII block. Now checked against the **union** of literal + symlink-resolved session-relative parts (keeps existing literal coverage, adds the resolved-target coverage). Fixed in **both** the Claude and Kimi copies. *(test: symlink alias → blocked, both adapters; read-guard 118/118.)*

### Resolved with justification
- **🟡 Codex-adapter raw capture has no hook layer** (`:494`). True, and my own source comment concedes it: Codex installs have no PreToolUse hook, so the opt-in `massread-evidence` capture relies there on the env gate (agent can't set it; operators enable it only for data they already hold) — the **primary masked-rail guarantee holds on every adapter**. The `:101` fix hardens the Claude/Kimi hook; a cross-adapter server-side redaction API is the v2 path. Documented bounded residual.
- **🟠 authn-vs-authz HIGH ceiling** (`claims.js:797`, `producer.js:782`). The same operator-decided point: a MEDIUM ceiling would permanently under-report genuine leaked-credential mass-reads in a fully-autonomous pipeline. The evaluator carries the under-privileged judgment under the loud operator contract; the signed booleans never claim `bfla_proven`. **v2's second-victim arm** is the tracked follow-up that makes it machine-proven. Re-logged, not re-litigated.

### Deferred — precision-safe recall gaps (miss leaks, never over-mint)
DOB/address value-shapes (`:383`), attacker-body-on-error (`:685`), redirect-to-login control recall (`:474`) — v1 boundaries.

Gates (local, full): `node --test` **55/0** (+5) · `test:mcp` **3052/0** · `test:hooks` **166/118/28** (+2 read-guard) · `test:prompts` **202/0** · `check:syntax` clean.

@codex review
@coderabbitai review

vmihalis:
## Round-6 (`dd3f377` → `3a156e4`) — disposition of the 7 threads

The counter reaches its converging model; two orthogonal real bugs are closed; the residual is the design ceiling.

### Fixed in `3a156e4`
- **Subject-count, converged** (`:419` P1, `:313` 🟡, and the whole #387/#393/#394 family). `distinct_pii_count` is now the count of distinct **SUBJECT KEYS** — *one key per record*, the sorted set of its normalized subject-identifier values (email/SSN/card/IBAN). One record → one key, however many PII tokens it carries, so a field listing two of a subject's own emails is one key not two (`:419`); same-bucket multi-field and a promoted nested self-collection collapse to one key (`:313`); phone + postal address are **excluded** from the key (a subject has several; phone format/country aliases over-count — `:183`) but stay **labeled** in `sensitive_field_names`. Deliberate precision-first v1 choice: a phone/address-only leak is recorded but doesn't mint a HIGH alone. *(tests: phone-only → no-mint; two-emails-one-record → one key.)*
- **🟡 IBAN target screen** (`:643`). `sensitiveShapesPresent` covers `detectPiiShapes` (email/phone/ssn/card) but not the IBAN shape this producer adds — so an IBAN in a path segment (`/accounts/GB82WEST…/transactions`) would persist into the signed target. Now screened explicitly at the target gate and the masked-summary gate. *(test: IBAN-in-path → blocked.)*
- **🟡 read-guard ordering** (`:202`, both Claude + Kimi). A blocked dir must DOMINATE the basename allowlist — a raw-PII file inside `massread-evidence/` whose basename matched `ALLOWED_EXACT` could slip past. Moved the `BLOCKED_DIRS` check (union of literal + symlink-resolved parts) **before** `ALLOWED_EXACT`. *(read-guard 118/118.)*

### Resolved with justification — the design ceiling (operator-accepted v1)
- **🟠 authn-vs-authz** (`:824`) and **🟠 endpoint not bound to a collection** (`:630`) are the **same** architectural truth: a schema-blind differential proves a credentialed caller bulk-read ≥2 distinct identifier-bearing subjects an anon client is denied — it does **not** machine-certify under-privilege, nor that the surface is a cross-subject collection vs a self-service endpoint. This is now stated honestly in the tool description + evaluator role as a **necessary-not-sufficient** signal: the evaluator (credential provenance + route kind) and grader **certify the class**, and the producer's HIGH is a *ceiling* the grader can lower — with the operator reading the report before disclosure as a final backstop. **The operator's informed call is to keep HIGH for v1** (a MEDIUM ceiling would permanently under-report genuine leaked-credential mass-reads in a fully-autonomous pipeline). These cannot be closed by code; **v2's second-victim arm + route-kind binding** is the tracked follow-up that makes the class machine-proven. Re-logged, not re-litigated.

### Deferred — precision-safe recall (miss leaks, never over-mint)
DOB/address value-shapes (`:383`), attacker-body-on-error (`:685`), redirect-to-login control recall (`:474`). And the Codex-adapter has no hook layer, so the opt-in `massread-evidence` capture relies there on the env gate + operator-only-for-owned-data (`:494`) — documented bounded residual; the primary masked rail is unaffected on every adapter.

Gates (local, full): `node --test` **58/0** (+3) · `test:mcp` 0-fail · `test:hooks` **166/118/28** · `test:prompts` **202/0** · `check:syntax` clean.

@codex review
@coderabbitai review

vmihalis:
## Round-7 (`3a156e4` → `b246034`) — disposition of the 7 threads

Three more real correctness/precision bugs fixed; the residual is the operator-accepted design ceiling. (The `test (20)` red on 3a156e4 was a **pre-existing flake** — an IDOR/reflect-producer test untouched here, green on the other 3 CI shards and locally 0-fail; this push re-runs CI.)

### Fixed in `b246034`
- **🟠 Shared-identifier over-count** (`:432` P1). The subject key per record is now its single **highest-priority** identifier (email > SSN > card > IBAN), not the full identifier set — so two rows that **share** that identifier are ONE subject even if one row carries an extra identifier (`{email}` vs `{email, ssn}`). That was the last same-subject inflation; `:419` (one record, many tokens → one key) stays closed. *(test: shared-email + extra-ssn → one subject.)*
- **🟡 Control reads one PII record but scored "denied"** (`:784`). A clean denial now requires the control surfaced **zero** subject PII. A 2xx control returning a single PII-bearing record (a public teaser / partial-public endpoint) was admitting a false HIGH; now `distinct_pii_count > 0` on the control → `control_inconclusive`, fail closed. *(test: control with 1 PII record → inconclusive.)*
- **🟠 IBAN case/encoding bypass** (`:650` P1). IBAN detection now **decodes** percent-encoding, scans **case-insensitively**, and validates the **mod-97 checksum** — so `gb82west…` / `%47%42…` are caught (the bare uppercase regex missed them), while hex-ish id/path segments are **not** false-positived (the bare shape regex would, and since the screen is fail-closed that would spuriously block legit runs). Applied at the body detector, the target screen, and the masked-summary screen. *(tests: lowercase IBAN → blocked; invalid-checksum hex path → mints, not false-blocked.)*

### Resolved with justification — the design ceiling (operator-accepted v1)
- **🟠 `:820` / 🟡 `:821`** (and `#824`/`#630` before them): the HIGH rests on an authenticated-vs-anonymous differential that does not by itself machine-certify under-privilege, nor that the surface is a cross-subject collection. This is now stated honestly as a **necessary-not-sufficient** signal in the tool description + evaluator role — the evaluator (credential provenance + route kind) and grader certify the class, the HIGH is a grader-lowerable **ceiling**, and the operator reads the report before disclosure. **Operator's informed call: keep HIGH for v1** (a MEDIUM ceiling would permanently under-report genuine leaked-credential mass-reads in a fully-autonomous pipeline). Not code-closeable; **v2's second-victim arm + route-kind binding** is the tracked follow-up. (`#821` itself notes the code is "admirably honest" about this.)

### Deferred
- **Cookie cross-host scope** (`:678` P1) — v1 sends the stored attacker cookies to the routed endpoint host; per-subdomain scope fidelity needs the jar to carry each cookie's captured domain/path → **v2 jar** (consistent with the round-3 disposition of `:208`).
- **Codex-adapter has no hook layer** (`:532`/`#494`) — the opt-in `massread-evidence` capture relies there on the env gate + operator-only-for-owned-data; documented bounded residual, the primary masked rail is unaffected on every adapter.
- Precision-safe recall: DOB/address shapes (`#383`), attacker-body-on-error (`#685`), redirect-control recall (`#474`).

Gates (local, full): `node --test` **62/0** (+4) · `test:mcp` **3059-pass 0-fail** · `check:syntax` clean (hooks/prompts unchanged this round).

@codex review
@coderabbitai review

vmihalis:
## Round-8 (`b246034` → `37ea5fe`) — disposition of the 8 threads

Four more real bugs fixed; the counter reaches its principled definition. (CI is green; the prior `test (20)` red was a confirmed flake — an IDOR/reflect-producer test untouched here, green on the other 3 shards and locally.)

### Fixed in `37ea5fe`
- **🟠 Overlapping same-shape over-count** (`:473` P1) — and with it the whole counter family. Distinct-subject counting is now **union-find over normalized subject-identifier values**: two records sharing ANY identifier value are the same subject. `{email:"a"}` vs `{email:"a alt"}`, `{email}` vs `{email,ssn}` (`#432`), multi-token-one-field (`#419`), same-bucket multi-field / promoted self-collections (`#393/#313`), boilerplate — all reduce to "records sharing an identifier" and merge. Failures are toward **under**-counting (safe). *(test: overlapping-alias → one subject.)*
- **🟡 Control with phone/address PII scored "denied"** (`:819` P1). A clean denial now requires the control surfaced **zero** PII of any kind (`pii_bearing_count`, not just `distinct_pii_count`) — a 2xx control returning one phone/address record was a false-HIGH path; now → `control_inconclusive`. *(test: control 1 phone record → inconclusive.)*
- **🟡 Formatted-IBAN bypass** (`:230` P1). The IBAN screen strips separators (spaces/hyphens) after percent-decoding, so `GB82%20WEST%20…` (decodes to `GB82 WEST …`) is caught; the mod-97 checksum still prevents false positives on hex-ish segments. *(test: %20-formatted IBAN in path → blocked.)*
- **🟠 Grep not guarded** (`:101` P1). `session-read-guard` now also covers the **Grep** tool (its `path` search root), on both the Claude and Kimi adapters + the source-tree settings — closing a path by which an agent could Grep the raw-PII `massread-evidence/` capture that Read/Bash already block. *(tests: Grep into blocked dir/file → blocked; non-session path → allowed; both adapters.)*

### Resolved with justification — the design ceiling (operator-accepted v1)
- **🟡 `:838`** (Claude, design): the dominant anon `302→/login` denial is scored `control_inconclusive`, so the common real case doesn't mint. Acknowledged — this is a **recall** limitation (the tool mints *less*, never a false HIGH); redirect-as-denial needs reliable opaqueredirect/login detection and is the tracked v2 follow-up alongside the victim-arm. Deferred, safe direction.
- The HIGH-ceiling threads remain the operator-accepted necessary-not-sufficient boundary (grader + operator-report-review backstop; v2 victim-arm + route-kind binding close it).

### Deferred
- Cookie cross-host scope (`:290` → v2 jar with per-cookie domain/path).
- Gmail dot/plus alias normalization (`:202`, P2) — minor precision edge; union-find already merges exact-normalized duplicates.

Gates (local, full): `node --test` **65/0** (+6) · `test:mcp` **3062-pass 0-fail** · `test:hooks` **166/124/28** (+6 read-guard) · `test:prompts` **202/0** · `check:syntax` clean.

@codex review
@coderabbitai review

vmihalis:
## `37ea5fe` → `0820c70` — v1 severity ceiling **HIGH → MEDIUM** (operator decision)

After the 8 review rounds, the operator made the call to lower the v1 ceiling — which resolves the recurring **design-ceiling** threads (`#820`/`#821`/`#824`/`#630`) at the root rather than merging over them.

**Why MEDIUM is the honest ceiling.** The signed evidence is an authenticated-vs-**anonymous** differential: it proves a credentialed caller bulk-reads PII an unauthenticated client is denied — a real exposure — but it does **not** machine-certify that the credential is *under-privileged* (a fully-authorized user reading authorized data also satisfies it) or that the surface is a cross-subject *collection*. That is honestly a MEDIUM. You were right to keep flagging the HIGH.

**What this buys:**
- the design-ceiling threads are **resolved**, not logged-over — the row no longer over-claims;
- the (now union-find-robust) subject counter's residual edges become minor MEDIUM over-counts the grader/operator absorb, not false criticals auto-flowing through an autonomous pipeline;
- the **HIGH (cross-tenant BFLA)** stays honest for the **v2 second-authenticated-victim arm** to *earn* with machine proof.

**Change:** `OFFENSIVE_TOOL_DEMONSTRATED_CEILING.bob_http_massread_confirm: "high" → "medium"` (authoritative registry) + the frozen producer mirror; `demonstrated_severity` is still registry-stamped, never agent-supplied. Tool description, producer header, evaluator role + regenerated agent/adapter surfaces restate "MEDIUM is the honest authn-differential ceiling; HIGH is v2." Tests updated (ceiling = medium; over-severity now rejects a `high` claim above the MEDIUM ceiling).

Gates: `node --test` **65/0** · `test:mcp` **3062-pass 0-fail** · `test:prompts` **202/0** · `check:syntax` clean.

This is intended as the converging change. @codex review @coderabbitai review

vmihalis:
## Round-8 review (`37ea5fe`) → addressed in `f53cfd0` (atop the MEDIUM flip)

These 5 landed during the HIGH→MEDIUM flip. Two are severity-independent (security); two tighten the now-MEDIUM oracle; all fixed.

- **🟡 Gmail-alias inflation** (`:203`). Email identifiers now normalize through the repo's Gmail-aware canonicalizer (`normalizeEmailForComparison`), so dot/plus aliases of one mailbox collapse to one value and union-find merges them into one subject. *(test: two Gmail aliases → one subject.)*
- **🟠 Double-encoded IBAN bypass** (`:224` P1). `findIbans` now decodes to a **fixed point** (bounded loop) before separator-stripping, so `%2520WEST…` is caught — a single decode pass left `%20` and the raw IBAN would have persisted into the signed target. *(test: double-encoded IBAN in path → blocked.)*
- **🟠 Singleton control PII** (`:821` P1, subsumes `:836`'s PII sub-case). Control denial now also rejects a **singleton-object** control body carrying field-bound PII (`controlBodyExposesPii`): `200 {"email":"teaser"}` or a 401 `{"error":..,"phone":..}` extracts no records but is real PII the anon client read → `control_inconclusive`. Field-bound, so a support email in denial free-text doesn't trip it. (`:836`'s residual — an unrecognized non-empty 2xx with *no* field-bound PII — is genuinely not the leaked data, so scoring it denied is correct.) *(test: singleton-object control with PII → inconclusive.)*
- **🟠 Raw-capture cross-target** (`:904` P1). The opt-in capture gate is now **bound to target**: `BOB_MASSREAD_OWNER_AUTHORIZED=<target_domain>` (CSV allowed); a bare `1` is rejected — it would have authorized raw-PII capture for every target the same MCP server evaluates. *(test: env naming a different domain → no capture.)*

**Context:** the v1 ceiling was lowered **HIGH → MEDIUM** (commit `0820c70`) — the differential proves authn-vs-anon bulk-read, honestly a MEDIUM; HIGH (cross-tenant BFLA) is the v2 victim-arm. That resolved the design-ceiling threads at the root; these 4 are the remaining correctness/security edges.

Gates: `node --test` **69/0** · `test:mcp` **3066-pass 0-fail** · `test:prompts` **202/0** · `check:syntax` clean.

@codex review
@coderabbitai review

vmihalis:
## Round-11 (`6cad1c6`) — all 5 unresolved threads on `f53cfd0` closed

CI was green and Codex returned **0 new comments**, but 5 threads remained — two flagged by **both Codex and brutalist** (strong signal). All five fixed with real code/doc changes, not waved off.

| # | Thread | Disposition |
|---|--------|-------------|
| **brutalist :188** | credit_card/IBAN as standalone subject identifiers can false-mint (`/me/payment-methods` = one person's two cards → 2 components → clears floor) | **Fixed.** Removed `credit_card` + `iban` from `SUBJECT_IDENTIFIER_SHAPES` (now `{email, ssn}`). They share the exact multiplicity property already used to exclude phone/postal address — a person holds several of each, so two card/IBAN *values* can't tell two subjects from one subject's two instruments. Still detected + recorded (`pii_shape_present`, `sensitive_field_names`), just don't **form** subject keys. A card/IBAN-only collection now under-counts (the safe direction); a real cross-subject financial leak almost always also carries a per-row email/SSN. *(tests: card-only no-mint, IBAN-only no-mint, reframed card-with-email mint.)* |
| **Codex P2 :837** | a 401 control with a non-JSON (CSV/text/HTML) body still bearing bulk PII is wrongly scored "denied" | **Fixed.** New `distinctSubjectsInText()` raw-text scan folds into `controlReadAnyPii` at a **>= MIN** threshold. Catches a genuine bulk export `deriveMaskedSummary`/`controlBodyExposesPii` can't parse; a lone support email (1 < MIN) does **not** trip it (preserves round-10's field-bound free-text robustness; structured single-subject PII is still caught at >= 1). *(tests: 401 CSV bulk → inconclusive; 401 lone support email → mints.)* |
| **Codex P1 :916** | on Codex (no hook layer) the opt-in raw capture is readable by the agent's own file tools | **Fixed.** New `readGuardActive()` — raw capture is now **default-DENIED** on adapters without a session read-guard. Requires BOTH `BOB_MASSREAD_OWNER_AUTHORIZED=<target>` AND a read-guard (env override `BOB_READ_GUARD`, else auto-detect of the `.claude`/`.kimi` read-guard hook file in the install tree). Codex/generic-mcp ship no such hook → no capture. Fails closed. **Producer-side only** — deliberately no `.mcp.json` env, which would break the `install-smoke` exact-shape `deepEqual` assertions across read-guarded *and* non-guarded adapters. Honest (like #131): incidental-read protection, not a same-UID-arbitrary-code boundary. *(tests: read-guard deny, read-guard auto-detect.)* |
| **Codex :41 + brutalist :41** | `evaluator-spawn.md` quick-reference still says `→ HIGH` and `BOB_MASSREAD_OWNER_AUTHORIZED=1` | **Fixed.** Quick-ref → `MEDIUM` + `=<target_domain>` + email/SSN-only subject-key wording; regenerated the codex/kimi/claude skill+agent copies; `prompt-contracts.test.js` comment `HIGH-ceiling` → `MEDIUM-ceiling`. The detailed `evaluator.md` bullet was already MEDIUM — this spawn quick-ref was the one surface missed during the HIGH→MEDIUM flip. |

**Gates:** `node --test` **75/0** · `test:mcp` **3072-pass 0-fail** · `test:prompts` **202/0** · `test:hooks` pass · `check:syntax` clean · NUL clean.

The v1 ceiling is MEDIUM (the differential proves authn-vs-anon bulk-read; cross-tenant BFLA = the deferred v2 victim-arm). With the subject model converged on union-find over 1:1 identifiers only, the control-denial oracle hardened against non-JSON bulk, and the raw capture default-denied on unguarded adapters, the remaining work is the tracked v2 follow-ups (second-victim arm for HIGH, redirect-as-denial recall, cookie cross-host jar).

@codex review
@coderabbitai review

vmihalis:
## Round-12 (`f5513a6`) — all 6 threads on `6cad1c6` closed

The round-11 review came back with 6 threads — **two critiqued the round-11 read-guard fix itself** (correctly), the rest are deeper oracle precision/recall issues the adversarial pass surfaced. All closed.

| # | Thread | Disposition |
|---|--------|-------------|
| **Codex P1 `:325` + brutalist `:324`** | read-guard auto-detect is defeated — the framework *ships* the hook file, so a Codex/mixed install still detects `true` | **Fixed.** `readGuardActive()` is now **explicit-only** (`BOB_READ_GUARD=1`), dropping the file-existence check. File presence ≠ enforced/registered hook. Default-DENY, fails closed. Both bots offered "require an explicit override" as acceptable. |
| **brutalist `:512`** | only top-level field names scanned → nested PII (`{user:{email}}`) missed entirely | **Fixed.** `deriveMaskedSummary` now walks each record's tree (bounded: depth ≤ 8, ≤ 512 fields/record). Masked-rail invariant preserved; serialize+recurse double-find is idempotent under the union-find/value-sets. The producer now works on ordinary nested API shapes. |
| **brutalist `:500`** | cross-row email/SSN over-counts → "always under-counts" claim is false | **Fixed (honest).** Corrected the overclaim. A schema-blind oracle cannot tell "N subjects" from "one subject's N instruments" for *any* value-based identifier (e.g. `/me/linked-emails`) — exactly why the differential is **necessary, not sufficient**, bounded by MEDIUM + operator contract + grader. SSN is 1:1; email is the collection key (can't remove). |
| **brutalist `:914`** | control scored "public" on record count alone → redacted-vs-PII mislabeled/dropped | **Fixed.** PII signal computed first; a control reading ≥ MIN records now splits: **with PII** → `control_also_reads_bulk_pii` (public, no mint); **redacted** → `control_inconclusive_redacted_bulk` (a field-level redacted-vs-PII differential v1 conservatively *declines* — the v2 oracle — no longer mislabeled "public"). Still never mints when the control read the collection (#669/#677 preserved). |
| **Codex P2 `:581`** | non-JSON control with bulk phone/card/IBAN (no email/SSN) ignored by the raw-text scan | **Fixed.** `distinctSubjectsInText` → `distinctPiiValuesInText` counts **all** PII shapes, so bulk non-identifier PII at a 401 forces inconclusive. ≥ MIN threshold preserved (lone support value still passes as clean denial). |

**Gates:** `node --test` **78/0** (+3) · `test:mcp` **3075-pass 0-fail** · `test:prompts` 0-fail · `test:hooks` 28/28 · `check:syntax` clean · NUL clean.

These were genuine — the read-guard auto-detect was a real hole in my own round-11 fix, and the nested-PII recursion materially improves the producer's reach on real APIs. The remaining residuals are the documented necessary-not-sufficient boundary (per-row identifier multiplicity, bounded by the MEDIUM ceiling + human grader) and the explicit v2 follow-ups (second-victim arm for HIGH cross-tenant proof; redacted-vs-PII field-level mint; cookie cross-host jar).

@codex review
@coderabbitai review

Issue #157, opened 2026-06-23
Mass-read producer v2: second-victim arm for HIGH (cross-tenant BFLA) + close v1 residuals


vmihalis:
## Mass-read producer v2 — earn HIGH (cross-tenant BFLA) + close the documented v1 residuals

`bob_http_massread_confirm` shipped in #156 at a **MEDIUM** ceiling: its attacker-vs-uncredentialed-control differential proves a credentialed caller bulk-reads PII an **anonymous** client is denied — a real exposure, but *not* machine-proof of under-privilege / cross-tenant access. The HIGH class and several recall gaps were deliberately deferred to keep v1 sound. This issue tracks v2.

### Primary: second authenticated "victim" arm → HIGH
Add a second **authenticated victim** identity arm. Mint HIGH only when: attacker reads the victim's records, the victim arm reads its OWN records (proving the data is the victim's private scope), and a third principal is denied — a categorical cross-principal break (mirrors the IDOR producer's canary-differential). This is the only sound path to a machine-certified cross-tenant BFLA HIGH; the registry ceiling lifts to `high` only behind this arm.

### Recall gaps consciously deferred in v1 (all SAFE-direction today: no false mint, but missed findings)
- **Redacted-vs-PII field-level differential** (#914): a control that reads ≥ MIN **redacted** records (no PII) while the attacker gets PII columns is currently `control_inconclusive_redacted_bulk` (declines). v2 could mint this field-level BFLA when route semantics confirm it's a shared collection.
- **Per-row identifier multiplicity over-count** (#500): a schema-blind union-find over email/SSN can over-count a per-row self-collection (`/me/linked-emails`). Bounded today by MEDIUM + grader + operator contract. v2: route-kind signal (collection vs self-service) from the evaluator to gate the mint.
- **Redirect-as-denial control recall** (#838/#474): a 3xx→login redirect control isn't scored a clean denial.
- **Cookie cross-host scope** (#290/#678): per-cookie domain/path jar so a cookie isn't sent cross-host.
- **POST-body listing + bearer-token identities**: v1 is GET-only + cookie-expressible identity (`blocked_by_design` otherwise).

### Integrity boundary (unchanged, documented)
The row MAC is tamper-evident against an MCP/guarded-Bash-confined agent but **not** un-forgeable by a same-UID actor (#131). Raw capture's read-guard is incidental-read protection, not a same-UID-arbitrary-code boundary. Both need the deferred offensive-sandbox (UID/container) work.

[...]

PR #159, opened 2026-06-24
feat(offensive): mass-read victim arm — cross-tenant BFLA HIGH (v2, #157)

vmihalis:
## What

Adds the **v2 victim arm** to `bob_http_massread_confirm` so it can machine-certify a **cross-tenant BFLA → HIGH**. v1 (#156) could only mint MEDIUM: the attacker-vs-anon differential proves a credentialed caller bulk-reads PII an *unauthenticated* client is denied — real, but not proof of under-privilege. This is the primary of #157.

## The oracle (opt-in)

Pass `victim_surface_id` (a routed in-scope surface for a **second** authenticated identity's OWN private scope, e.g. `/api/me`) + `victim_auth_profile` (default `"victim"`). HIGH is stamped **iff all four hold**:

1. the **victim arm** 2xx-reads ≥1 of its own subject identifiers;
2. a **fresh anon** client is **DENIED** that same victim scope (so it is private);
3. the **attacker's bulk read OVERLAPS** one of those identifiers (computed value-blind in memory — only the **count + booleans** are recorded; the masked rail never carries a raw value);
4. attacker and victim are **distinct** (distinct profile name + cookie jar).

The arm is **elevating only**: any failure leaves the proven MEDIUM intact and records why (`cross_tenant_proven` + `victim_elevation`). It runs **after** the base differential passes, so a failing base never pays for the extra two arms.

## Severity wiring — the soundness core

`offensive-capture-writer.js`'s `demonstratedSeverityOverride` can only **lower** the registry ceiling (`Math.max` over a descending `SEVERITY_VALUES`). So to mint HIGH the **registry ceiling rises `medium → high`** (the max a row *may* demonstrate), and the producer **always** passes an explicit override — `"medium"` for v1, `"high"` only on a proven break, **never `undefined`** (which would fail *open* to the high ceiling).

This is safe because severity is capped **twice, independently** (`claims.js` `assertExploitedClaimHasProof`):
- **Gate 1** — claim ≤ `max(row.demonstrated_severity)`: keyed on the **row's actual** demonstrated_severity, so a claim citing a v1 MEDIUM row is capped at MEDIUM *regardless of the registry*.
- **Gate 2** — `row.demonstrated_severity ≤ registry`: raising it to `high` merely *permits* a proven row to demonstrate high.

⇒ **A v1 MEDIUM row can never back a HIGH claim, even though the registry now permits high** (asserted by an inverted ceiling test).

## Honest residual (documented in the tool description + evaluator brief)

Distinct sessions don't prove distinct *humans*, so principal-distinctness rests on the **operator contract** (a synthetic Bob-minted victim, or an owner-controlled second account) + the 3-round blind verification + grader. The signed row records the mechanical legs (`victim_read_own_private_scope` / `victim_scope_anon_denied` / `attacker_read_victim_subject` / `victim_distinct_session`) so a reviewer sees exactly what was proven. Hardening this with a synthetic-provenance hard gate is a noted follow-up.

## Safety properties preserved

- **Masked rail**: the signed row still carries only counts / field-name buckets / booleans — never raw PII. The overlap is an in-memory intersection over normalized identifiers; only the count survives. Fail-closed PII screen unchanged.
- **Opt-in raw capture**: victim + anon-victim bodies are added to the capture only under the existing `BOB_MASSREAD_OWNER_AUTHORIZED=<target_domain>` **and** `BOB_READ_GUARD` gates.
- **v1 byte-identical**: the control-denial predicate is factored (`controlReadsAnyPii` / `controlIsCleanDenial`) so the v1 listing control and the v2 anon-victim control score denial through **one** path (no drift between two copies of a security check); all 78 existing tests pass unchanged.

## Tests & gates

- 11 new victim-arm tests: HIGH happy-path, nested victim identifier, HIGH claim round-trip, and **every** elevation-decline path (no overlap, anon-not-denied, same profile, same cookie jar, victim profile absent, victim-arm WAF, unroutable victim surface, victim absent). **89/89** in the producer file.
- Full `npm test` green end-to-end (`test:mcp` 3086, `test:prompts` 202, `test:hooks` 28, `test:install`/`cli`/`adapter`/`package`/`policy-replay` all pass), `check:syntax` + NUL scan clean. Regenerated `.claude/agents/*` + codex/kimi adapter skills.

Closes the #157 primary (HIGH victim arm). v1 recall residuals (#914 field-level, #500 multiplicity, redirect-as-denial, cookie cross-host, POST/bearer) remain tracked in #157.

## Summary by CodeRabbit

* **New Features**
  * Extended the `bob_http_massread_confirm` mass-read confirmation tool with optional victim context inputs to support cross-tenant BFLA verification.
  * Updated severity outcomes: MEDIUM for the standard attacker-vs-control differential; HIGH when the cross-tenant victim-arm oracle is proven.
* **Documentation**
  * Refreshed tool, evaluator, and signed-producer contracts to describe the new optional parameters and revised MEDIUM vs HIGH rules.
* **Security/Validation**
  * Added a fail-closed signing guard to prevent elevated severity being recorded without an explicit per-run decision.
* **Tests**
  * Expanded cross-tenant/victim-arm coverage, including MEDIUM downgrades and negative cases for incorrect eligibility and severity signing.


vmihalis:
@codex review

vmihalis:
@coderabbitai review

vmihalis:
Thanks — all legitimate. The common root cause across the high-severity findings was that the victim arm **inferred ownership from the response body**. `e083083` re-anchors the proof to the victim's **known identity** (the synthetic Bob-minted profile's account email), never response contents. Per finding:

- **Session distinctness (#A)** — replaced `cookieJarsEqual` (full-set equality, bypassable by a shared `sid` + any incidental cookie) with `cookiesShareAnyValue`: a **disjoint-jar** requirement that fails closed on a shared session cookie.
- **Response-inferred ownership (#B/#C/#D/#E)** — the victim must now be a **synthetic** profile carrying a known email (`victimIdentityKey`, mirroring the IDOR `#18` gate). The overlap is anchored to **that** email, so a multi-subject listing, org/team view, replayed collection, or shared `support@` boilerplate can't forge ownership. Also: the victim arm must read its **own** known identity from the scope, and `victim_surface_id` must **differ** from the bulk listing.
- **Anon leak in unstructured text (#F)** — `subjectIdentifiersInRawText` scans the anon denial body's raw text, so a `401` like `login required for victim@x` is not scored "private".
- **Severity fail-open (#G)** — `buildAndSignOffensiveRow` gains `requireExplicitSeverity`; the producer sets it, so any future path that omits the override **fails closed** instead of signing the high ceiling.

Plus a same-human guard (when the attacker is also synthetic, its email must differ from the victim's).

**Tests:** 95/95 in the producer file (6 new covering exactly these paths + a sign-guard unit test); IDOR 110/110 and capture-writer 8/8 unaffected; full `npm test` green. Tool description + evaluator prompts updated for the synthetic-victim contract.

@codex review
@coderabbitai review

vmihalis:
Round 2 — all 4 addressed in `a783961`:

- **#H (alias replay)** — the surface-id check only caught an identical id. Now compares the **resolved canonical targets**: a different `victim_surface_id` that routes to the same listing endpoint declines (`victim_endpoint_equals_listing`).
- **#I (unauditable HIGH)** — `victim_surface_id` + the (PII-screened, value-blind) victim canonical target now fold into the **hash-bound stderr bundle**, so the signed row MAC-covers which victim scope established the break. New assertion in the HIGH test verifies it.
- **#J (cross-name session reuse)** — `cookiesShareAnyValue` now compares token-shaped **values** (len ≥ 16) regardless of cookie name, so the same opaque session token under a different name declines. It deliberately does **not** decline on a shared *short* benign cookie (`locale`/`consent`/`role`) — that would gut recall and isn't evidence of a shared session (test added for both directions).
- **#K (stale integrity ceiling)** — the tool description + producer header now state the same-UID forge residual as **HIGH** (raised from MEDIUM with the ceiling), closeable only by the deferred offensive-sandbox.

**98/98** producer tests (+3), full `npm test` green.

@codex review
@coderabbitai review

vmihalis:
Round 3 — addressed in `28a1503`. **#L (multi-subject victim scope):** the victim's own-scope read must now be **single-subject** (exactly one distinct subject identifier — the victim's own `/me` record). A private team/org/members listing that merely *includes* the victim now declines (`victim_scope_multi_subject`) and stays MEDIUM, so `victim_read_own_private_scope` is only signed when the scope is genuinely the victim's own. Conservative on purpose (a `/me` embedding a second person's email also declines — the safe direction).

99/99 producer tests (+1 multi-subject), full `npm test` green.

@codex review
@coderabbitai review

PR #160, opened 2026-06-24
fix(offensive): mass-read v2 single-subject gate counts records, not identifier keys (live-validation finding)

vmihalis:
## What

Fixes a soundness bug in the v2 mass-read victim arm (#159) that the **first live firing** surfaced — a real `/me` carrying one person's **email + SSN** was wrongly rejected as multi-subject, so the cross-tenant HIGH never fired.

## How it was found

I stood up a cookie-session validation lab (`/Users/memehalis/sec/offensive-lab/`) and drove the **real producer** (live Patchright browser, `authed_fetch`, real `Set-Cookie` sessions) against it — the first end-to-end firing, not a seeded driver. v1 MEDIUM fired clean; v2 fell back to MEDIUM with `victim_elevation: "victim_scope_multi_subject"`.

## Root cause

The #L single-subject gate used `subjectIdentifierSet(victim.body).size === 1`, which counts distinct identifier **keys**. One person's `/me` legitimately carries several identifier *shapes* (`email` **and** `ssn`) → 2 keys → falsely "multi-subject". Every real `/me` with email+SSN (KYC / financial / gov) would have failed HIGH. All 99 unit tests put only `email` in the victim body (size 1 either way), so they missed it.

## Fix

Count **records** (people), not identifier keys — a new `recordCountOf(body)` (a collection's element count, or 1 for a singleton `/me` object). One record = one subject regardless of identifier shapes. The #L intent is preserved: a multi-subject org/team listing is a multi-**record** collection → count > 1 → still declines.

## Verification

- **Re-fired live after the fix:** v1 MEDIUM **and** v2 HIGH both fire (`cross_tenant_proven:true`, `overlap:1`) and mint MAC-signed `offensive-runs` rows; the masked rail carries no raw PII against real bodies.
- Regression test added (a `/me` with email+ssn+phone for one person → HIGH). **100/100** producer tests, `test:mcp` green, `check:syntax` clean.

Follow-up to #159 / #157.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved v2 victim-elevation “single-subject” detection by traversing the full request payload, counting distinct normalized subject identifiers (including multi-shape cases), and failing closed when traversal limits are reached.
  * Updated attacker/victim overlap logic to enforce that overlap is derived from bulk data records (not response/relational metadata) and anchored to matching victim subject leaf keys with varying-path evidence.

* **Tests**
  * Added v2 “victim-arm” regression coverage for single-person multi-identifier eligibility, multi-subject rejection, and record-bound attacker overlap enforcement.

### DIFF
--- a/mcp/lib/offensive-massread-producer.js
+++ b/mcp/lib/offensive-massread-producer.js
@@ -641,7 +641,8 @@ function subjectIdentifierSet(bodyText) {
   // extractRecords-selected subset: a /me's OWN top-level identifier must be seen even when the body ALSO
   // embeds a recognized child collection (`{email, items:[...]}` — extractRecords would pick `items` and miss
   // the top-level email, bot-review #N), and a wrapper / object-map's nested subjects must all be counted
-  // (#M / CodeRabbit). Used in the victim arm for membership (does X contain victimKey) + distinctSubjectCount.
+  // (#M / CodeRabbit). Used in the victim arm for the own-identity + anon-leak MEMBERSHIP tests only (does this
+  // body contain victimKey); the single-subject COUNT and the attacker OVERLAP each use their own walk below.
   let nodeBudget = MASSREAD_MAX_NODES_PER_RECORD * MASSREAD_PII_SCAN_RECORDS; // whole-body, bounded
   const visit = (node, depth) => {
     if (node == null || typeof node !== "object" || depth > MASSREAD_MAX_RECORD_DEPTH || nodeBudget <= 0) return;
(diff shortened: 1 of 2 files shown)


vmihalis:
@codex review

vmihalis:
@coderabbitai review

vmihalis:
Good catches — `recordCountOf` was the wrong tool too. `af2f430` switches the gate to **`distinctSubjectCount(body) === 1`** = the MAX over each subject-identifier shape (email/ssn) of its distinct normalized values anywhere in the body. Shape-agnostic to the container:

- **#M (wrapper `{members:[…]}`, unrecognized key)** → 2 distinct emails → rejected.
- **CodeRabbit (object-map `{user1:{…},user2:{…}}`)** → 2 distinct emails → rejected.
- **#N (`{email, items:[…]}`, recognized child collection)** → the child items carry no subject IDs → 1 email → allowed.

I also fixed `subjectIdentifierSet` to scan the **whole body** instead of an `extractRecords`-selected subset — otherwise `{email, items:[…]}` made `extractRecords` pick `items` and miss the top-level email (so `victimReadsOwnIdentity` went false). Only the victim arm's scans use it.

Documented residual (exotic, safe-leaning): two people with **disjoint** shapes and ≤1 of each (email-only + ssn-only subjects) reads as 1 — bounded by the operator pointing `victim_surface_id` at a real single-subject `/me` + the 3-round verification + grader.

**103/103** producer tests (+3: wrapper, object-map, self-owned child collection; email+ssn regression retained), `test:mcp` green.

@codex review
@coderabbitai review

vmihalis:
Both P1s are real — fixed in `52936ee` with a principled **subject model** instead of shape-counting.

**P1 #1 (mixed-shape multi-subject):** `distinctSubjectCount` is now **union-find with object-node grouping** — identifiers that are DIRECT SCALAR field values of the **same object node** are one subject (a `/me`'s `{email, ssn}` siblings); identifiers in **different** object nodes are different subjects unless they share a value. So `{email:victim}` + `{ssn:other}` as separate nodes → **2** → rejected. SCALAR-only direct extraction is deliberate: a sensitively-named container key whose value is an object/array (`{contacts:[{email:a},{email:b}]}`) must not serialize-and-merge its children into the parent — those children are their own nodes, reached by recursion (that was the false-HIGH vector). One /me `{email,ssn}` → 1; array / unrecognized wrapper (#M) / object-map → 2+; self-owned child collection (#N) → 1; two disjoint-shape people → 2. ✓ all cases.

**P1 #2 (overlap not record-bound):** new `recordBoundSubjectIdentifierSet` scans **only the extracted bulk records**, so `attackerReadsVictim` requires the victim to appear as a **row the attacker read**, not echoed in metadata like `requested_by_email`. `victimReadsOwnIdentity` / `anonSeesVictimIdentity` keep the whole-body scan (membership tests — a /me's own top-level email and any anon leak must be seen anywhere; safe direction).

**Residual (documented, SAFE-leaning → at worst a missed HIGH that stays MEDIUM, never a false HIGH):** one subject split across nested objects (`{email, profile:{ssn}}`) reads as 2, and a structured-field / array-of-scalars identifier is not counted as a node's own subject — both err toward rejecting the single-subject gate (fall back to v1 MEDIUM).

**Tests:** 105/105 producer (+2: mixed-shape → MEDIUM; record-bound metadata overlap → MEDIUM). Flat email+ssn one-person `/me` → HIGH and nested `{profile:{contact:{email}}}` → HIGH both still pass; #M / object-map / #N unchanged. `test:mcp` 3102/0, `check:syntax` clean.

@codex review
@coderabbitai review

vmihalis:
All three are real — same root cause (a sensitively-named field can name a **different principal** than the record's subject). Fixed in `d69273f` with a principled model.

**Count — `distinctSubjectCount` (lines 752 + 743):** now per **object node** = MAX over subject-shape of the count of **distinct normalized values** among that node's sensitive fields, **summed across nodes**.
- **#752** two people in sibling fields `{email:victim, teammate_email:other}` — the round-2 union collapsed both same-shape ids to one subject; max-per-shape now counts **2 distinct emails in the node → 2 → rejected**.
- **#743** another subject under a non-scalar sensitive field `{email:victim, secondary_email:[other]}` — scalar-only skipped the array; values are now **serialize-extracted**, so 2 → rejected. Under max-per-shape, serializing a container is **safe** (a multi-subject container reads as ≥2 → MEDIUM, never a false-HIGH 1). One person's `{email,ssn}` stays **1 → HIGH**.

**Overlap — `attackerReadVictimSubject` (lines 696/699):** record-bound wasn't **subject**-bound. Now bound to the row's own subject via **cross-record variance**, no field-name parsing: per field NAME, a field whose values **vary** across records (≥2 distinct) is a SUBJECT field; a field **constant** across records is shared metadata (one viewer/requester stamped on every row). The victim overlaps iff `victimKey` is a value of a **varying** field — so `{email:other1, viewer_email:victim}` no longer signs HIGH (`viewer_email` is constant metadata; `email` varies and doesn't hold the victim). The base gate already requires ≥ MIN distinct subjects, so a genuine victim-subject is still caught.

**Residual (documented, SAFE-leaning → at worst a missed HIGH that stays MEDIUM, never a false HIGH):** one subject split across separate nested objects, or carrying multiple distinct same-shape values (primary+recovery email), reads as 2 → MEDIUM. The single err-toward-HIGH case (two *different* people's *different* shapes fused into one flat object, `{email:alice, ssn:bobs}`) isn't a realistic API record and is bounded by the synthetic-victim anchor + 3-round verification + grader.

**Tests:** 108/108 producer (+3 targeting exactly #752/#743/#696). Flat email+ssn one-person `/me` → HIGH and nested `{profile:{contact:{email}}}` → HIGH still pass; #M / object-map / #N and the round-2 cases unchanged. `test:mcp` green (the lone local failure was a `browser_driver_ready_timeout` flake in the Patchright transport test — unrelated to this pure-logic diff; passes 35/35 in isolation), `check:syntax` clean.

@codex review
@coderabbitai review

vmihalis:
Four of the five are real and fixed in `aa7e7df`; the fifth (Codex #760 cross-shape) is the **irreducible subject-inference ambiguity** — I've bounded and logged it rather than chase a denylist. Detail:

**Fixed**
- **CodeRabbit #760 (budget) — `distinctSubjectCount` fail-closed.** If the node budget is exhausted mid-walk it returned the partial `total`; a large object-map scanned only up to the victim node could read as `1`. Now returns `≥ 2` on exhaustion (a partial walk can never pass the single-subject gate).
- **Codex #702 (nested metadata path).** The variance map keyed on the bare leaf name merged `viewer.email` with the row subject `email`. Now keyed by **full path** → distinct.
- **Codex #711 (varying metadata).** Variance alone treated *any* ≥2-distinct field as a subject, so `manager_email`/`referred_by_email` varying per row could sign HIGH on a victim-as-manager. Added **field-consistency**: the matching path's leaf key must be one under which the victim's **own `/me`** carried its identity (`victimSubjectLeafNames`). Metadata under a different key can no longer pose as the subject.
- **CodeRabbit #1703 (test).** Strengthened so `viewer_email` is truly **constant across records** (2 rows, same value) + a third row keeps base distinct subjects ≥ MIN.

**Logged residual — Codex #760 cross-shape siblings (not fixed; bounded):**
`{email: victim, spouse_ssn: other}` → `max(1 email, 1 ssn) = 1` → accepted as single-subject. This is genuinely undecidable from response shape: a bare `ssn` sibling is **indistinguishable from the victim's own ssn** — which is *exactly the live-validation case this PR exists to support* (a real `/me` carries the caller's `email`+`ssn` → must stay HIGH). Distinguishing `ssn` from `spouse_ssn` requires field-name semantics, i.e. an incomplete relational-qualifier denylist that invites endless rounds on each new qualifier.

It's bounded, soundly:
1. The residual requires the victim's **own `/me`** to embed **another real principal's** identifier in a sibling field. The victim is a **synthetic, freshly-minted `bob_auto_signup` account** — it has no spouse/dependent/teammate data, so its `/me` cannot carry another person's `ssn`.
2. Otherwise gated by the operator pointing `victim_surface_id` at a real single-subject `/me`, the 3-round blind verification, and the grader.
3. The robust future upgrade is **canary-injection** (write a unique nonce into the victim's own record and find it in the attacker's read, like the IDOR producer) — unambiguous, no subject inference. Tracking as a follow-up; out of scope here.

I'm making the gate-exception call to keep this residual rather than add a denylist, per the merge-rule "you judge each finding." Same-shape multi-subject, sensitive containers, nested/varying/constant metadata are all now caught.

**Tests:** 109/109 producer (+1 varying-metadata `manager_email`; `viewer_email` strengthened in place). Flat `email+ssn` one-person `/me` → HIGH and nested `{profile:{contact:{email}}}` → HIGH still pass; all prior-round cases unchanged. `test:mcp` green, `check:syntax` clean.

@codex review
@coderabbitai review

vmihalis:
## Merge decision — round-5 findings logged as a bounded class (gate-exception)

The three round-5 Codex P1s are all the **same irreducible class**: the victim's email appearing in the attacker's bulk read via a **non-subject path** —
- #748 nested relational path (`manager.email`),
- #690 self-echoed audit field in `/me` (`created_by_email`),
- #739 co-mingled multi-value field (`email:"other, victim"`).

Per the merge rule ("you judge each finding; CodeRabbit/Codex is sometimes wrong — you may merge with a justification logged on the PR"), I'm merging with these three logged as **bounded, not reachable in the actual flow**, for one structural reason:

**The v2 HIGH only ever mints for a Bob-minted SYNTHETIC victim.** `demonstrated_severity: "high"` requires a non-null `victimKey` (`offensive-massread-producer.js:1285,1293`), and `victimIdentityKey` returns null unless `profile.synthetic === true` (`:870`) — a flag set **only** by the in-process `bob_auto_signup` path. So the victim is always a fresh account Bob just created on the target, which **has no relationships in the target app**: it is no one's manager, created no one's record, is co-listed with no one. Every round-5 (and round-4 #760) vector requires exactly such a relationship for the victim's email to appear as relational/audit/co-mingled metadata in the attacker's bulk read of *other* accounts. Those responses cannot occur for a fresh synthetic victim — the only way its email appears in the bulk is as **its own leaked record** (the cross-tenant break we certify).

The bots review `attackerReadVictimSubject` in isolation and can't see this enforced invariant, so they correctly flag the shapes — but the shapes are unreachable given how the victim is provisioned.

**What IS fixed in this PR (real bugs, kept):** the single-subject count now uses per-object-node max-per-shape (one person's `{email,ssn}` → 1, multi-subject containers → ≥2), sensitive non-scalar containers are counted, the overlap is subject-bound via cross-record variance + field-consistency with the victim's own `/me` key, and the count fails closed on budget exhaustion. These are defense-in-depth that also harden the non-synthetic/misconfiguration cases.

**Why not keep iterating:** no *read-only* overlap escapes the invariant — any victim identifier can in principle appear as metadata. The robust escape is **canary-injection** (write a unique nonce into the victim's own record, like the IDOR producer, and require that exact nonce in the attacker's read — no subject inference). That's a read+write feature, tracked as a dedicated follow-up, not rushed into this fix PR.

CI green (Bob Diff Review, brutalist, CI ×4, CodeRabbit — all SUCCESS); 109/109 producer + test:mcp 3106/0. Merging.

Issue #161, opened 2026-06-24
Mass-read v2 HIGH: replace email-overlap with canary-injection (escape the synthetic-victim invariant)


vmihalis:
## Problem

The v2 mass-read victim arm (`bob_http_massread_confirm`, #157/#159/#160) certifies a cross-tenant BFLA at **HIGH** by proving the attacker's bulk read contains the synthetic victim's identity. The proof is an **email/identity OVERLAP**. Four rounds of adversarial review (#160) established that an email-overlap is *fundamentally* insufficient to prove the attacker read the victim's **record**, because the victim's email can appear in the attacker's response via non-subject paths:

- relational fields (`manager_email`, `manager.email`, `viewer_email`)
- audit fields (`created_by_email`, `requested_by_email`)
- co-mingled multi-value fields (`email: "other, victim"`)
- cross-shape siblings (`{email: victim, spouse_ssn: other}`)

#160 hardened the heuristics (per-node max-per-shape count, cross-record variance, field-consistency, budget fail-closed) and these vectors are currently **bounded by the structurally-enforced synthetic-victim invariant** — HIGH only mints for a `profile.synthetic===true` victim (`offensive-massread-producer.js:870,1285`), a fresh `bob_auto_signup` account that has **no relationships** in the target app, so its email cannot appear as relational/audit/co-mingled metadata in a bulk read of *other* accounts. That argument is sound, but it is an **invariant the reviewer can't see in the overlap function**, and no read-only heuristic escapes it.

## Robust fix — canary injection (mirrors the IDOR producer)

Replace (or augment) the email-overlap with a **unique canary** the victim controls:

1. When provisioning the synthetic victim, write a high-entropy nonce into a victim-controlled, readable profile field (e.g. `display_name` / `bio` / a custom attribute) via an authenticated write. (Requires the producer to gain a bounded write step — currently read-only — gated like the IDOR producer's canary write and the `BOB_HTTP_WRITE_CONFIRM` HITL hook.)
2. Confirm the canary is in the victim's own `/me`.
3. The cross-tenant HIGH fires iff that **exact nonce** appears in the attacker's bulk read.

A random nonce in a victim-controlled field cannot be relational/audit metadata about the victim on someone else's row, is not co-mingled, and needs no subject inference — so the HIGH no longer relies on the synthetic-victim invariant for soundness; it becomes a direct proof the attacker read the victim's record.

## Considerations
[...]

PR #162, opened 2026-06-24
feat(offensive): bob_http_idor_confirm live arm — operator-armed canary self-provisioning (PR-D)

vmihalis:
## What

Activates the **live arm** of `bob_http_idor_confirm` (PR-D) — Bob's first **write-capable** offensive producer. The 7-probe P0–P8 cross-tenant **canary** oracle + MAC signing were already built but INERT (the dispatcher called `idorConfirm` with no `provision`). This adds the live self-provisioning that feeds the oracle.

## How

- **`liveProvision()`** — creates O_A/O_B/O_C IN ORDER via authenticated POST (so a creation-order-dependent broken read-grant lands the A→B break), mints 3 distinct 256-bit canaries, captures the server-minted ids, reads B's object back AS B (canary-leaf discovery + #24 create-time PII screen). `createObject()` honors the same injectable `fetchFn` as the probes — unit tests drive the full arm with one mock; a real run uses `safeFetch`.
- **Create-contract** (metadata only; AC-3 still forbids the canary VALUE + object ids): `canary_field`, `id_field` (default `"id"`), optional **synthetic** `create_body`. The create endpoint is **derived** from the AC-2-bound `path_template` (drop trailing `/{id}` → the collection) — structurally on-route, **no new off-route write surface**. `create_body` is PII-screened **before any write**.

## Operator checkpoint (the live write gate)

- **INERT by default.** Live creates fire ONLY when the operator arms THIS target: `BOB_IDOR_PROVISION_AUTHORIZED=<target_domain>` (target-bound, never a bare `1`). Unarmed → `blocked_by_design:object_not_self_provisioned`, no network. `global_preapproval:false` keeps per-call approval; the #18 provenance gate still requires three synthetic Bob-minted identities. Create POSTs are the only writes — scope-validated + audited.
- `demonstrated_severity` is a **HARD MEDIUM** by construction (attacker-owned synthetic objects); #13/#14 weak-attribution absence demotes to LOW via confidence signals (cap-not-kill), never inflates.

## Validation

- **115 producer tests** — 110 existing seeded-path unchanged + 5 PR-D: env target-binding, unarmed→inert, `create_body` PII→blocked-before-write, `liveProvision` create-order/capture/readback, and the **full armed create→readback→P0-P8→signed MEDIUM mint** (stateful mock). `test:mcp` green (a lone `browser_driver_ready_timeout` was Chrome-launch contention from a concurrent live fire — passes 16/16 in isolation). `check:syntax` clean.
- **LIVE FIRE** against a local cookie-session lab (real HTTP, real `Set-Cookie` auth via a `Cookie`-header profile, real create POSTs + read probes + MAC signing): `confirmed:true`, `demonstrated_severity:medium`, **zero confidence signals**, a MAC-signed `offensive-runs` row, no identity mailbox in the signed row.

## Boundaries / scope

Integrity boundary unchanged (#131 in-process key: tamper-evident vs an MCP/guarded-Bash agent, not un-forgeable by a same-UID actor). v1: direct path-segment-id reads, single-endpoint single-host, single-collection POST + single canary field. Ties #161 (the mass-read canary reuses this provisioning machinery).

## Summary by CodeRabbit

* **New Features**
  * Added operator-armed live self-provisioning for `bob_http_idor_confirm` (A→B→C), gated by `BOB_IDOR_PROVISION_AUTHORIZED`, using server-minted IDs plus canary verification.
* **Bug Fixes**
  * Unarmed calls now fail closed as `blocked_by_design` before any network activity; only `GET` is permitted (explicitly blocking `HEAD`).
  * Strengthened request/path and create-collection write-surface validation, including tighter baseline binding and stricter truncation/foreign-data handling with clearer blocked reasons.
* **Documentation**
  * Updated tool and evaluator guidance for the arming workflow and `differential_response` evidence wiring.
* **Tests**
  * Expanded live arming, hardening, and leakage/denial coverage.


vmihalis:
@codex review
@coderabbitai review

vmihalis:
All 11 are legitimate (the gate correctly hammering the new **write surface**) — fixed in the latest commit.

**Write-byte screening (CR #888, Codex #887/#701/#895)** — the single biggest cluster. The pre-write PII/secret gate now screens **every written byte** — the create URL, the serialized body, **and** the canary field NAME — with **layered decoding** (percent/unicode to a fixed point), mirroring the proof-body gates. One consolidated scan over `${createUrl}\n${body}\n${field}` + its decoded form → `blocked_operator_pii:create_inputs_contain_sensitive_value`. So an encoded `victim%40example.test`, a token in `canary_field`, or a sensitive segment in the derived URL can no longer be written.

**Codex #893 (inert-suffix create-path)** — `createPath` now strips the **final path segment** (`/[^/]*$`), so `/api/accounts/{id}.json` → `/api/accounts` instead of POSTing a malformed `%7Bid%7D.json`.

**Codex #760 (fail-fast)** — `liveProvision` validates each create's captured id **before** the next POST, so a bad contract / wrong `id_field` no longer triggers the later B/C target writes.

**CR #15 (docs)** — the tool description now states the MEDIUM **ceiling** downshifts to LOW via `confidence_signals` when cross-tenant attribution is unprovable (cap-not-kill).

**Tests (CR #2144/#2183/#2204/#2249)** — the mock now **enforces** the declared `canary_field` (the `liveProvision` test uses a non-default field end-to-end); the unarmed test asserts **zero create POSTs**; the create-inputs test covers raw PII, **percent-encoded** PII, an **AWS secret** shape, and **PII in the field name** (each asserting zero POSTs); the masked-rail check rejects **all three** identity mailboxes + their token markers.

**Re-validated:** 115 producer tests, `test:mcp` green, `check:syntax` clean, and the **live lab fire still mints a signed MEDIUM** (confirmed, MAC-signed, no identity PII) after the create-path + screening changes.

@codex review
@coderabbitai review

vmihalis:
All 6 are legitimate — fixed in `1ef6894`.

**Captured-id handling (Codex #771/#744×2):** a created id is interpolated into the readback/probe URLs, so it's now validated at capture, before any use.
- `createObject` only accepts an id from a **2xx** create whose `id_field` is an **own scalar** (`hasOwnProperty` + string/number) — a 4xx/5xx error envelope with an `id`/request-id no longer reads as a created object (#744a), and `toString`/`constructor` no longer resolves to a prototype function treated as an id (#744b). Both restore fail-fast on a bad contract.
- `liveProvision` validates each captured id (`capturedIdSegmentIsSafe` + layered-decoded PII/secret) **before** the owner-readback GET dispatches, so a `foo/transfer` id can't trigger a credentialed GET to `/api/.../foo%2Ftransfer` ahead of `idorConfirm`'s id checks (#771).

**Codex #870 (per-call gate) — you're right, verified.** `permissionsForAllTools()` (merge-claude-config.js:382) allow-lists every shipped tool, so `global_preapproval:false` does **not** create a per-call prompt. I've removed that overclaim from the description + checkpoint comment. The honest, real gate: the **target-bound arming env** (`BOB_IDOR_PROVISION_AUTHORIZED`, default-off, out-of-band of a confined MCP/Bash agent — the same boundary as the row-MAC key #131, mirroring the opt-in `BOB_HTTP_WRITE_CONFIRM` flag #135) + the #18 synthetic-provenance gate + the every-written-byte PII/secret screen. I kept `global_preapproval:false` (scope-bound session-authority, not `preapproval_global`) but no longer claim it prompts. Design call (logged): the env-arming opt-in is Bob's established write-gate pattern; I'm not adding a per-call-prompt hook, consistent with #135.

**Tests (CR #2155/#2207):** the readback mock matches the **exact** `/api/accounts/{id}` route (not `endsWith`) so a target-binding regression can't pass; new case proves PII in a fixed path segment flows into the derived `createUrl` and is refused before any write (postCount 0).

**Re-validated:** 116 producer tests, `check:syntax` clean, and the **live lab fire still mints a signed MEDIUM** (confirmed, MAC-signed, no identity PII) after the id-capture + screening changes.

@codex review
@coderabbitai review

vmihalis:
All 5 are legitimate write-surface edge cases — fixed in `646f875`. (Also fixed the CI regression below.)

**P1 — never fire a real mutation**
- **Action-shaped create collection (909):** correct, and the most important one. `assertReadOnlyPath` intentionally allows action-*nouns* like `transfer`/`refund` for a GET-by-id (reading a transfer record is safe), but the derived collection POST `/api/transfer` would *execute* the action. New `assertCreateCollectionShapeSafe` holds the derived create path to a stricter, fail-closed action-verb set (financial / lifecycle / destructive / auth / execution verbs) than the read path — each segment decoded to a fixed point and split on `. _ -` so `/transfers`, `/transfer-funds`, `/transfer_v2` all surface the `transfer` action. Over-blocking is the deliberate bias for a write primitive; the only cost is LIVE auto-provisioning is unavailable on an action-shaped endpoint (the seeded path still works).
- **HEAD (930):** correct. HEAD has no body → the canary differential can never mint, yet an armed HEAD call would still self-provision A/B/C (3 live writes) before the body-less probes block. The handler now refuses GET-only **before** any provisioning, and the schema narrows `method` to `["GET"]`.

**P2 — request integrity / unsafe inputs**
- **Unsafe numeric id (751):** correct. A 64-bit id outside `Number.isSafeInteger` was already rounded by `JSON.parse`, so the readback/probe URL would address a *different* resource. `createObject` now accepts a number only when `Number.isSafeInteger`; strings always OK; anything else → "no id captured" → fail-fast.
- **Uncapped create body (702):** correct — `safeFetch` caps responses, not request bodies. The serialized `create_body` is now capped at 8 KB before any write.
- **Prototype-pollution canary field (900):** correct. `__proto__`/`constructor`/`prototype` as `canary_field` / `id_field` / any `create_body` key are refused before any write (`create_body` keys arrive via `JSON.parse`, which defines `__proto__` as an *own* key, so `Object.keys` sees it).

**CI regression (test 82, "Claude adapter config never leaks into the neutral MCP runtime"):** the round-2 honesty edit named `merge-claude-config.js` in the neutral tool descriptor, whose `claude-config` substring trips the module-boundary-hygiene guard. Rephrased to keep the honest content (the installer's generated permission allow-list pre-approves shipped tools → `global_preapproval:false` is not a per-call prompt) without naming the adapter config file — which a neutral runtime descriptor shouldn't do anyway.

**Re-validated:** 121 producer tests, `test:mcp` 3117/0-fail, `test:prompts` 202/202 (leak test green), `check:syntax` clean, and the **live lab fire still mints a signed MEDIUM** (confirmed, MAC-signed, no identity PII, `confidence_signals: []`) after the id-capture + guard changes. +5 round-3 tests, each asserting **zero create POSTs** when the guard trips.

@codex review
@coderabbitai review

vmihalis:
All 9 are legitimate — fixed in `dedb997`. CI is green again (round-3 leak regression resolved). This round moved from obvious write risks to subtler subversion of the create-CONTRACT metadata; good catches.

**Action-shaped create collection — bypass closes**
- **camelCase (Codex 226 / CR 77):** `/api/transferFunds/{id}` lowercased + split-on-`._-` tokenized whole and passed. The guard now breaks camelCase / PascalCase (`lower|digit→Upper` and `ACRONYM→Word`) before tokenizing, so `transferFunds` / `refundOrder` surface their verb the same as `transfer-funds`.
- **`order` (CR 77):** the doc claimed `/api/orders` was blocked but `order` wasn't in the verb set. Added it (a `POST /api/orders` places an order — a real mutation), so doc and behavior agree.

**create_body — recursively screened before any write**
- **nested `__proto__` (Codex 935):** top-level-only check missed `{"nested":{"__proto__":…}}`. `screenCreateBody` now walks objects + arrays (depth-bounded) and fails closed on a reserved key at any depth.
- **method/action dispatch (Codex 965):** `{"_method":"DELETE"}` / `{"action":"run"}` passed the PII-only scan; a target honoring body-level override could mutate despite the path guard. Reserved a focused override-key set (`_method/method/action/_action/op/operation/cmd/command/do`) at any depth.
- **client-supplied id (Codex 710):** `createBody` is spread unchanged into the POST, so `{"id":"known-object"}` (or whatever `id_field` names) could drive upsert-on-create to an attacker-chosen id, violating the server-minted-id invariant. The `id_field` key is now refused at the top level.

**Oracle-soundness + write-target**
- **canary_field overlapping a scope key (Codex 926):** the most important one — the producer writes a *distinct* minted canary per identity, and the oracle reads `owner_scope/tenant_id/org_id/workspace_id` as the private tenant discriminator (#13/#14), so a scope-key `canary_field` could forge "provably distinct tenants" from attacker-chosen values and strip the confidence downgrade, signing an inflated proof. Refused.
- **final-body cap (Codex 943):** the 8 KB cap measured only `JSON.stringify(createBody)`, but the POST body also includes the caller-supplied `canary_field` key + 64-hex value. Now caps the *final* serialized body incl. the field name.
- **root-level derived endpoint (Codex 954):** `/{id}` strips to an empty collection → `POST /` (site root, often login/contact/search), with no segment for the action guard. Refused.

**Test (CR 2390):** added the literal `prototype` reserved case.

**Re-validated:** 125 producer tests (+`prototype` case, +4 round-4 tests incl. an 8-case subversion matrix, each asserting **zero create POSTs**), `test:mcp` 3112/0-fail (11 patchright-skips are environmental), `test:prompts` 202/202, `check:syntax` clean, and the **live lab fire still mints a signed MEDIUM** (confirmed, MAC-signed, no identity PII).

@codex review
@coderabbitai review

vmihalis:
All 4 (3 distinct fixes) are legitimate — fixed in `1c673d3`.

- **`canary_field` aliasing `id_field` (Codex 739 + CR 2487):** correct — `createObject` writes the minted canary into `canary_field` *after* `screenCreateBody`, so `canary_field === id_field` (default `id_field` is `"id"`) puts the server-minted canary into the **object-id slot**. On a create/upsert API honoring client ids that POSTs an object with a client-supplied id, violating the server-minted-id invariant before the post-write `canary_reflected_in_object_id` guard can fire. Now refused before any write (`canary_field_aliases_id_field`; covers `canary_field:"id"` by default).
- **Matrix suffix (Codex 234, P1):** correct — `/api/transfer;v=1` (and an encoded `%3b`) tokenized whole because the split only handled `. _ -` + camelCase, so routers that strip the suffix would dispatch `POST /api/transfer`. The guard now also splits on matrix/param punctuation `; , : =` (the path is decoded to a fixed point, so the encoded form is caught). While here I also fixed a **latent defect from the round-4 edit**: the camelCase sentinel + split char had been stored as NUL bytes — functionally self-consistent but an invisible landmine (broke `grep`); rewritten to a visible `-` sentinel, zero NUL bytes.
- **`create_body` schema (Codex 39):** correct and important for the real-target run — the dispatcher validator treats an object schema without `additionalProperties` as **closed** (`tool-validation.js` defaults to `false`), so `create_body:{"name":"x"}` was rejected before the handler, making the documented skeleton unusable on APIs requiring non-canary create fields. Set `additionalProperties:true`; the handler's recursive pre-write screen (PII/secret, reserved/override/client-id keys, size cap) polices the content.

**Re-validated:** 129 producer tests (+canary/id-alias e2e, +a direct `assertCreateCollectionShapeSafe` unit test for camelCase / `;v=1` / `%3b` / commerce, +a `create_body` additionalProperties schema test via the real validator, +an armed-run e2e proving a synthetic non-canary field flows through and still mints MEDIUM), `test:mcp` 3116/0-fail (11 patchright-skips environmental; one reflect-producer host-binding flake under parallel load cleared on re-run), `test:prompts` 202/202, `check:syntax` clean, **live lab fire still mints a signed MEDIUM**.

@codex review
@coderabbitai review

vmihalis:
All 5 are legitimate — fixed in `b9ec29d`. This round is alias/normalization bypasses of the round-4/5 field-name guards, plus the workflow-integration gap.

**Alias / normalization bypasses (Codex P1)** — new `normalizeFieldName()` collapses camel/snake/kebab/spacing to one canonical form, applied to both guards:
- **`canary_field` scope alias (971):** the exact snake_case check missed `canary_field:"tenantId"` / `"org-id"` on APIs that normalize field names → the canary would land in the owning-scope field and forge "distinct tenants". Now rejected by normalized match against the owning-scope keys.
- **`create_body` client-id alias (146):** the screen blocked only the exact `id_field`; `{object_id:…}` / `{resourceId:…}` bypassed. `screenCreateBody` now rejects a normalized client-id alias set + the normalized `id_field` at **any depth**.
- **`create_body` owning-scope field (147):** `{tenant_id:"victim-org"}` (or a nested envelope) → scope drift during the write. Now rejected — any normalized owning-scope key at **any depth** (`create_body_scope_field`).

**Workflow integration (Codex descriptor:11):** correct and important — the descriptor flipped the tool to operator-armed but the evaluator playbooks still said INERT / "use manual replay", so real runs would skip the producer. Updated the **source** roles (`prompts/roles/evaluator.md`, `evaluator-spawn.md`) to describe the armed arm (env `BOB_IDOR_PROVISION_AUTHORIZED` + 3 synthetic identities → self-provisions canary objects + mints a signed cross-tenant MEDIUM via `differential_response`; unarmed → fall back to victim replay), and **regenerated** the codex / kimi / claude adapter surfaces (`generate-codex-skills.js`, `generate-kimi-roles.js`, `generate-claude-roles.js`).

**Test (CR 2564):** correct — the round-5 synthetic-field test would pass even if `create_body` were dropped. The mock now records every POST body; the test asserts `name`/`title` actually arrive on each create (canary still overrides). +a round-6 alias matrix (each asserting **zero create POSTs**).

**Re-validated:** 130 producer tests, `test:mcp` 3117/0-fail (11 patchright-skips environmental), `test:prompts` 202/202 (all generated skill/role surfaces in sync), `check:syntax` clean, **live lab fire still mints a signed MEDIUM**.

@codex review
@coderabbitai review

vmihalis:
Legitimate — fixed in `929ca48`. Round 6 added `bob_http_idor_confirm` to the mint-crediting `safe_oracle.kind: "differential_response"` mapping in the **long** evaluator role, but the **parallel line in the spawn role** (`prompts/roles/evaluator-spawn.md`) still listed only `bob_http_massread_confirm` — so a spawned evaluator citing a live-arm IDOR row would have a stale claim contract and `bob_record_candidate_claim` could reject / under-credit it. Added `bob_http_idor_confirm` to the spawn mint mapping and regenerated the claude / codex / kimi adapter surfaces. `test:prompts` 202/202 (all generated surfaces in sync).

@codex review
@coderabbitai review

vmihalis:
All 3 P1s are legitimate — fixed in `e34207d`.

- **canary_field aliasing the id field (1006):** round 5 rejected the exact `canary_field === id_field`, but an ID alias (`uuid`/`object_id`/`resourceId` or a case variant of `id`) still passed → `createObject` POSTs `{ [canaryField]: canary }`, so an upsert API normalizing the alias would treat Bob's canary as a client-supplied object id. The canary_field id check now rejects the **normalized** form against `ID_ALIAS_KEYS` + the normalized `id_field` (reusing round 6's `normalizeFieldName`).
- **URL-valued create_body fields (760):** `{avatar_url:"http://169.254.169.254/…"}` / `{callback_url:…}` was spread unchanged (the prior screen was PII/secret only) → SSRF / off-target callback through the canary create. `screenCreateBody` now refuses a URL-shaped string value (`scheme://` or protocol-relative `//`) at **any depth** (`create_body_url_value`); anchored so a benign `"note: hi"` / `"ratio 3:1"` is not a URL.
- **Audit preflight before the create POST (771):** the mutating create fired *before* `auditConfirmRequest`, so an audit-write failure left the mutation outside the circuit-breaker / budget trail. `createObject` now **reserves** a request-initiated audit record (status null) before the POST and aborts if it can't be written — no unaudited mutation escapes; the completion record (with status) is still appended after.

**Re-validated:** 133 producer tests (+canary id-alias matrix, +URL-value matrix, +an audit-preflight test proving the POST never fires when the audit can't be recorded — each asserting **zero create POSTs**), `test:mcp` 3120/0-fail, `test:prompts` 202/202, `check:syntax` clean, **live lab fire still mints a signed MEDIUM**.

@codex review
@coderabbitai review

vmihalis:
All 4 are legitimate — fixed in `196ef00`. (The one CI failure on the prior commit was the known `offensive-reflect` host-binding flake — real HTTP under parallel load, a file this PR doesn't touch; it didn't recur locally and the new commit re-runs CI fresh.)

- **URL strings in create_body ARRAYS (160, P1):** `screenCreateBody` checked URL values only on object properties, so `{callbacks:["http://169.254.169.254/…"]}` slipped (the array branch recursed into the string element, which the primitive branch returned `null` for). Refactored to screen **string values at the top of the recursion**, so URL + GraphQL checks now cover object-property values *and* array elements at any depth.
- **GraphQL mutation payloads (127, P1):** on a GraphQL-compatible derived collection, `{query:"mutation { … }"}` would execute an arbitrary mutation. Now refused (`create_body_graphql_operation`); anchored on the operation keyword + optional name + `{`/`(`, so a benign `"mutation rate: 0.5"` is **not** matched (the test asserts it still mints).
- **Malformed-escape masking the path guard (224, P1):** `/api/transfer%3bv=%zz` left the whole segment raw (`decodeURIComponent` abandons it), hiding the `;`. `assertCreateCollectionShapeSafe` now decodes each **valid** percent-triplet to a fixed point (leaving malformed ones) and splits on a recovered `/` too.
- **Raw-readback foreign PII (887, P2):** the #24 gate scanned only the *parsed* readback, so PII in bytes `JSON.parse` discards (a shadowed duplicate key) was missed. `liveProvision` now captures the raw readback bytes and #24 scans them too.

**Re-validated:** 137 producer tests (+array-URL, +GraphQL-op incl. benign-word-still-mints, +malformed-escape path, +raw-readback-PII — each fail-closed), `test:mcp` 3124/0-fail, `test:prompts` 202/202, `check:syntax` clean, **live lab fire still mints a signed MEDIUM**.

@codex review
@coderabbitai review

vmihalis:
All 7 are legitimate — fixed in `44a3b6d`.

- **Owner-assignment fields (186, P1):** the scope screen only blocked `owner_scope`/`tenant_id`/`org_id`/`workspace_id`; `user_id`/`owner_id`/`account_id`/`created_by`/`team_id` (etc.) were POSTed, letting a create API assign the synthetic object to a caller-chosen real principal. New normalized `OWNER_ID_KEYS`, refused at any depth (`create_body_owner_field`), ordered after the id-field check so the configured id still reports the precise `client_supplied_id`.
- **Non-slash HTTP schemes (143, P1):** `http:host` / `http:/host` / `http:\host` (WHATWG-normalized to absolute URLs) slipped the `scheme://`-only regex. Now matches known fetch schemes with optional/back-slashes + a host char, plus `//`/`\\`, while still not matching `"file: report"` / `"note: hi"` / `"ratio 3:1"`.
- **GraphQL ignored tokens (148, P1):** `mutation,{…}` and `mutation #c\n {…}` bypassed the whitespace-only detector (GraphQL treats commas + `#` comments as ignored). The detector now allows ignored tokens before an optional name and the `{`/`(` — `"mutation rate: 0.5"` still not matched.
- **Embedded-id template (1078, P1):** `normalizePathTemplate` forbids anything *after* `{id}` but not a prefix in the same segment, so `/api/account-{id}` / `/api/v1/accounts.{id}` stripped to `/api` / `/api/v1` (off-route POST). The live arm now requires the final segment to be exactly `{id}` (optionally `{id}.ext}`) before deriving (`create_collection_not_id_terminated`).
- **Truncated readback (899, P2):** the readback was trusted for canary discovery + #24 even when `safeFetch` flagged truncation. `liveProvision` now stamps `owner_readback_b_truncated` and `idorConfirm` fails closed, mirroring the proof-body non-truncation gate.
- **Decode + secret-scan raw readback (1227, P2):** the round-9 raw scan ran only `piiScan` on literal bytes. Now scans raw **plus** `decodeAllEncodingLayers(...)` through `piiScan` **and** `secretShapesIn` — an encoded `victim%40corp.test` or shadowed `sk-`/`AKIA` token is caught.
- **Test (CR 2745):** the raw-readback-PII test now also asserts **zero** offensive rows signed.

**Re-validated:** 143 producer tests (+round-10 matrices: owner fields, truncated readback, encoded/secret raw readback, embedded-id templates, non-slash URLs, GraphQL ignored tokens — each fail-closed, zero create POSTs/no row), `test:mcp` 3130/0-fail, `test:prompts` 202/202, `check:syntax` clean, **live lab fire still mints a signed MEDIUM**.

@codex review
@coderabbitai review

vmihalis:
All 6 are legitimate — fixed in `36b4d46`. Rather than enumerate more aliases, this round **closes the bypass classes** the prior rounds kept spawning variants of:

- **Owner/scope relationship fields → base-noun matching (201/146):** bare `tenant`/`org`/`workspace` and bare `user`/`team`/`created_by_id` slipped the `*_id`-only sets. Replaced the enumerated sets with `SCOPE_BASE_NOUNS`/`OWNER_BASE_NOUNS` matched via `baseNoun()` (normalize, then strip a trailing `id`/`by` twice), so the bare form, the `_id` form, and the `_by[_id]` form all collapse to one base — refused at any depth. A benign field ending in a stripped suffix (`created_at`, `title`) is **not** over-blocked (test asserts it still mints).
- **URL values → WHATWG parse (154):** `http:[::1]` / `http:@169.254.169.254` (bracketed / userinfo authorities) slipped the alnum-authority regex. `valueLooksLikeUrl` now layers protocol-relative + a broadened explicit-scheme regex (any authority char incl. `[`/`@`) + a `new URL()` parse whose protocol is a fetch scheme — catching every spelling Node normalizes to an absolute URL, while `"file: report"`/`"note: hi"`/`"ratio 3:1"` stay clean.
- **GraphQL → keyword + selection set (160):** `mutation @custom { … }` (directives) and comment/comma variants bypassed the structured regex. Now flags a `mutation`/`subscription` keyword co-occurring with a `{` — closing the grammar tail in one rule; `"mutation rate: 0.5"` (no `{`) still mints.
- **Decoded backslash separator (255):** `/api/%5Ctransfer` → `\transfer`; added `\` to the path-split class so a framework decoding `%5C` as a separator can't hide the action verb.
- **Test (CR 2796):** the encoded/secret raw-readback test now also asserts **zero** offensive rows signed.

**Re-validated:** 147 producer tests (+round-11 matrices, incl. a benign-suffix-still-mints case), `test:mcp` 3134/0-fail, `test:prompts` 202/202, `check:syntax` clean, **live lab fire still mints a signed MEDIUM**.

@codex review
@coderabbitai review

vmihalis:
All 6 are legitimate — fixed in `1ed5aac`. These close consistency gaps in the round-11 class-closing screens (apply the same normalization/decoding everywhere), rather than new classes.

- **Encoded URL/GraphQL values (205):** the string screen ran the detectors on the raw value only; now also runs them over `decodeAllEncodingLayers(value)`, so an encoded `http%3a%2f%2f169…` is caught.
- **`canary_field` bare scope alias (1079):** the scope-overlap check used the `_id`-form set; a bare `tenant`/`org`/`workspace` (which some APIs normalize to `tenant_id` in the stored object) passed. Switched to the same base-noun match used for create_body keys.
- **Plural relationship suffixes (150):** `baseNoun` only stripped singular `id`/`by`; `tenant_ids`/`user_ids`/`teamIds` slipped. It now also strips a trailing `ids`/`bys` and a plural `s` — the plural forms collapse to the base noun (`created_at`/`title` still not over-blocked).
- **Method-override key normalization (218):** the override set was matched by exact lowercase, missing `method_override`/`methodOverride`/`X-HTTP-Method-Override`. Now normalized + matched via `normalizeFieldName`, with the aliases added.
- **Dotted/bracketed prototype keys (217, P2):** a deep-setter endpoint expands `constructor.prototype.x`/`__proto__[x]` into a prototype write; the reserved check now splits each key on `. [ ]` and matches every segment.
- **Create-URL host binding (1143):** `assertSingleEndpointSingleHost` already rejects a surface resolving to >1 origin (so a relative endpoint + a *different* declared host can't reach the bind — the specific scenario isn't reachable). As forward-safe defense-in-depth I still capture that single validated origin and fail closed unless `baselineUrl.origin` equals it, so no future change to `resolveBaselineFromSurface`'s origin ordering can point a live WRITE at the session apex.

Removed the now-redundant `NORMALIZED_SCOPE_KEYS` (base-noun supersedes it).

**Re-validated:** 150 producer tests (+round-12 matrices), `test:mcp` 3137/0-fail, `test:prompts` 202/202, `check:syntax` clean, **live lab fire still mints a signed MEDIUM**.

@codex review
@coderabbitai review

vmihalis:
All 4 are legitimate — fixed in `2c6104a`. Further consistency closure (apply the create_body treatment to `canary_field` + extend suffix coverage).

- **Dotted/bracketed reserved `canary_field` (1096):** `canary_field` becomes a JSON key, but its reserved check was exact-only — `constructor.prototype.polluted` / `__proto__[polluted]` reached the target. Factored `keyHasReservedSegment` (splits on `. [ ]`, matches every raw segment) and applied it to **both** create_body keys and the `canary_field`/`id_field` names.
- **UUID/GUID/UID scope+owner suffixes (160):** `baseNoun` stripped `id`/`by`/plural-`s` but not `uuid`/`guid`/`uid`, so `tenant_uuid`/`workspaceGuid`/`owner_uid` passed. `baseNoun` now also strips a trailing `uuid`/`guid`/`uid` (longest first; length-guarded so a **bare** `uid`/`uuid`/`guid` stays intact).
- **`uid` as an object-id alias (143):** bare `uid` normalizes to `uid` (not in the id set) and `baseNoun` reduced it to `u`, so `create_body:{uid}` / `canary_field:"uid"` slipped. Added `uid` to `ID_ALIAS_KEYS` → reports `client_supplied_id` / `canary_field_aliases_id_field`.
- **Test (CR 2900):** added `confirmed === false` beside the reason/zero-POST assertions in the URL + canary-scope cases.

**Re-validated:** 152 producer tests (+round-13 matrices), `test:mcp` 3139/0-fail, `test:prompts` 202/202, `check:syntax` clean, **live lab fire still mints a signed MEDIUM**.

@codex review
@coderabbitai review

vmihalis:
Fixed in `1001f72`. The brutalist's key insight closed a whole class, so I switched the field-name matching from whole-name normalization to **tokenization**.

- **Tokenized scope/owner selectors (brutalist prefix-bearing + Codex composite/slug, P1):** whole-name `baseNoun` collapsed `victim_tenant_id` → `victimtenant`, `owner_user_id` → `owneruser`; `tenant_slug`/`workspaceKey`/`org_code` slipped. `fieldIsScopeSelector`/`fieldIsOwnerSelector` now tokenize (split on non-alnum + camelCase) and match **any** token (plus a whole-name match for joined nouns like `owner_scope`) — prefixed/composite/slug forms all caught. The OWNER set is trimmed to clear selectors (content-ambiguous `created`/`creator`/`author` removed); the `*_by` actor pattern is caught by a `by`-token rule, so `created_at`/`updated_at` are **not** over-blocked (test asserts they still mint).
- **`canary_field` owner selectors (brutalist 842 / Codex 1117, P1):** `canary_field` is POSTed as a key but was only scope/id-checked — `user_id`/`owner_id`/`createdBy` would write the canary into an ownership slot. Now refused (`canary_field_overlaps_owner_field`), after the id-alias check.
- **Scheme-agnostic URL (brutalist 176 / Codex 143, P1):** `valueLooksLikeUrl` was a scheme enumeration. Replaced with protocol-relative + any `scheme://` + known no-slash fetch schemes + a WHATWG parse yielding a non-empty host — so `redis://`/`mongodb://`/`ssh://`/`jdbc:mysql://` are all caught, while `"file: report"`/`"note: hi"` stay clean.
- **Query-routed endpoint (Codex 1181, P1):** a query-qualified recorded endpoint (`/api/items/123?type=invoice`) is query-routed; the create is now refused (`create_collection_query_routed_endpoint`) rather than POSTing to the queryless `/api/items`.
- **Deep encoding (brutalist 229):** `decodeValidPercentTriplets` bumped 8→12 (with the 8 in `decodePathSegments`, ~20 layers), and the create guard fails closed on a residual triplet.

**On the design finding (brutalist 842 / Codex 64 — env-arming, no per-call prompt):** this is deliberate and documented, not an oversight. The target-bound arming env (`BOB_IDOR_PROVISION_AUTHORIZED`) is the operator's write-authorization gate; it mirrors the opt-in `BOB_HTTP_WRITE_CONFIRM` (#135) and shares the row-MAC key's out-of-band boundary (#131), and the descriptor states honestly that `global_preapproval:false` is *not* a per-call prompt. The create-contract screening is **defense-in-depth on top of** that operator gate — it bounds what a confined agent can do *with* an operator-armed target; it is not the primary boundary. I'm not changing the gate model in this PR (a per-call prompt would contradict the established #135 pattern). Logging this as a reviewed, intentional trade-off.

**Re-validated:** 156 producer tests (+round-14 matrices, incl. timestamp-still-mints), `test:mcp` 3143/0-fail, `test:prompts` 202/202, `check:syntax` clean, **live lab fire still mints a signed MEDIUM**.

@codex review
@coderabbitai review

vmihalis:
Fixed in `6343bc4`. All eight threads addressed — six concrete fixes plus the logged design call on the action-verb denylist.

- **Deep id-alias create_body keys (Codex P1, idor:275):** `id.uuid` / `uid[value]` collapse to a non-alias whole-name (`iduuid`/`uidvalue`) and bypassed the id-alias check, while `keyHasReservedSegment` already splits dotted/bracketed for proto keys. New `keyHasIdAliasSegment` splits the same shapes and matches each segment against `ID_ALIAS_KEYS` / the configured `id_field` — applied to create_body keys **and** the `canary_field` NAME.
- **Privilege fields (Codex P1, idor:270):** `{is_admin:true}` / `{roles:["admin"]}` / `{granted_scopes}` passed the screen, so an armed agent could mint an *elevated* synthetic object. New `fieldIsPrivilegeSelector` (tokenized like the scope/owner selectors; unambiguous markers only — role/admin/permission/privilege/grant/acl/scope/capability/authority/entitlement/superuser/sudo) refuses them (`create_body_privilege_field`), mirrored for `canary_field`. (`userRole`-style names are already caught one check earlier as `owner_field` via the `user` token — both block.)
- **GraphQL query body (brutalist/Codex, idor:225):** the screen was mutation-only, so `{"query":"query {...}"}` + a derived `POST /graphql` became an arbitrary authed GraphQL surface. `valueLooksLikeGraphqlMutation` → `…Operation` (matches `query` too), **and** `graphql ∈ WRITE_ACTION_VERBS` so the derived `/graphql` collection is independently refused by the action-shape guard.
- **Create-response screen (brutalist, idor:940):** the A/B/C create *responses* were never foreign-PII/secret-screened. `liveProvision` now scans each create response (raw + decoded) and fails closed via a tagged error → `create_response_contains_foreign_data` — parity with the #24 readback gate. (The first create POST necessarily fires before its response can be screened; B/C then never fire — exactly 1 POST, asserted.)
- **Nested real container (brutalist, idor:1215):** the stripped parent path may be a *fixed real* tenant container (`/api/orgs/acme/projects/proj-123/accounts`). `pathHasConcreteParentInstance` refuses the create when an ancestor segment is a concrete resource instance (numeric / uuid / id-bearing slug like `proj-123`). Residual (a pure-alpha tenant slug, indistinguishable from a route word) is documented and backstopped by the downstream same-tenant guard (`identities_collided_same_tenant` / `own_scope_missing`) so cross-tenant *soundness* still holds.
- **Residual percent-escape (Codex P1, common:249):** `%64elete%zz` decoded to `delete%zz`, which the valid-`%XX` residual check skipped and the tokenizer kept whole (`%` is not a delimiter), slipping the action guard. `assertCreateCollectionShapeSafe` now fails closed on **any** residual `%` after the valid-triplet decode.

**On the action-verb denylist (brutalist common:61 / Codex common:75 — "invert to an inert-noun allowlist"):** I **materially expanded** the set with the whole class the brutalist + Codex named (redeem/process/fulfill/dispatch/issue/escalate/assign/schedule/provision/deploy/convert/swap/stake/subscribe/renew/upgrade/notify/broadcast/invite/share/email/sms/webhook/password/graphql). I'm **not inverting to an allowlist** in this PR, and that is a deliberate, logged call: an *inert-REST-noun allowlist* is infeasible because it cannot enumerate product-specific resource nouns (widgets/gizmos/foobars/…), so it would either reject every custom-resource collection (making the live arm useless) or degrade to the same open-vocabulary problem. The honest architecture is that **no static classifier can decide "this POST has no side effects"** — so this denylist is **defense-in-depth on top of** the operator's target-bound write-arming env (`BOB_IDOR_PROVISION_AUTHORIZED`), and the operator arming a *specific authorized target* + the synthetic canary remain the primary boundary (mirrors `BOB_HTTP_WRITE_CONFIRM` #135 and the row-MAC key #131). Over-blocking is the correct bias — the only cost is that auto-provisioning is unavailable on an action-shaped endpoint. Logging this as a reviewed, intentional trade-off.

**Plus a latent fix #4 surfaced:** a chance 16-digit run inside a 64-hex canary false-positives as a credit card (~0.4%/canary) and would block a clean mint. `neutralizeCanaries` strips the producer's *own* minted canaries (exact 64-char run — cannot hide adjacent real PII) before the foreign-data scan, applied consistently to #4 (create responses), #24 (owner readback), and #17/#17b (proof bodies). 15× flake loop on the producer suite: 0 failures.

**Re-validated:** 161 producer tests (+round-15 matrices), `test:mcp` 3150/0-fail, `test:prompts` 202/202, `check:syntax` clean, **live lab fire still mints a signed MEDIUM** (MAC-valid, no identity PII, `confidence_signals` empty).

@codex review
@coderabbitai review

vmihalis:
Fixed in `570b50e`. The structural / parser-consistency gaps are now closed; the residual is the logged denylist-completeness design point.

- **Encoded + single-digit parent container (brutalist/Codex P1, idor:939 / idor:925):** `pathHasConcreteParentInstance` ran on the raw `createPath`, so `/api/users/%34%32/notes` (→ `42`) and `/api/orgs/org-1/accounts` (single digit) bypassed the nested-container guard while a router decodes them to a real fixed container. It now **decodes percent-encoding to a fixed point before splitting** (parser-consistent with the action-verb guard), re-splits on a recovered separator, **fails closed on a residual `%`**, and flags a single-digit separated id (`\d{2,}` → `\d+`).
- **Anonymous GraphQL shorthand (brutalist/Codex P1, idor:258):** the screen required a `query`/`mutation`/`subscription` keyword, so `{ viewer { email } }` (a valid anonymous query doc) slipped — and on a `/api` or `/gql` GraphQL endpoint the action-verb guard doesn't catch, it became an authed GraphQL read on the first POST. A brace-wrapped value that is **not valid JSON** is now treated as a GraphQL document (a JSON-object string still mints).
- **Credential field NAMES (brutalist/Codex P1, idor:316):** `screenCreateBody` screened value *shapes* but not credential field *names*, so `{password:"weak"}` / `{api_key:"local-test"}` whose values match no token regex were POSTed verbatim — contradicting the "every written byte secret-screened" contract. New `fieldIsCredentialSelector` refuses them (`create_body_credential_field`), mirrored for `canary_field`. It matches generic `secret`/`token`/`authorization` only as a **whole name** and `password`/`apikey`/`clientsecret`/… as bare tokens, so `secret_note` / `csrf_token` stay clean (regression-tested).
- **`company_id` tenant scope (Codex P1, idor:169):** added `company`/`enterprise`/`business` to `SCOPE_BASE_NOUNS` — `company_id` is an owning-tenant selector like `org_id`.
- **Truncated create response (Codex P1, idor:1059):** a `bodyTruncated` create response was screened only over the retained prefix, then its id accepted — foreign PII past the cap unseen. Now fails closed (`create_response_truncated`), mirroring the owner-readback / proof-body truncation gates.
- **CodeRabbit test hardening:** the contaminated + truncated create-response tests now assert **zero offensive rows are signed**; the `pathHasConcreteParentInstance` test covers encoded (`%34%32` / `proj%2d123`) and residual-`%` parents.

**On the two remaining design points:**

- **Action-verb denylist (brutalist idor:65 — "missing verify/confirm/accept/…"):** I added the named verbs (verify/confirm/accept/decline/complete/release/promote/demote/resend/deliver/merge). But the recurring "the denylist can't be complete" point is, at this stage, the **logged defense-in-depth design call**: the denylist is *not* the primary boundary — the operator arming a *specific authorized target* (`BOB_IDOR_PROVISION_AUTHORIZED`) + the synthetic canary are. An inert-noun *allowlist* is infeasible (it can't enumerate product-specific resource nouns: widgets/gizmos/…) so it would make the live arm useless on every custom resource. I'm not inverting; over-blocking is the correct bias (auto-provisioning is simply unavailable on an action-shaped endpoint).
- **Pure-alpha tenant slug residual (brutalist idor:920):** acknowledged — and the brutalist itself rates it "a self-documented limitation, not a missed bug." A slug with no id portion (`acme`, `mobile`) is statically indistinguishable from a route word; cross-tenant **soundness** is backstopped by the downstream same-tenant guard, and a fully operator-seeded known-synthetic parent contract is the deferred robust closure (ties to canary-injection #161).

**Re-validated:** 169 producer tests (+round-16 matrices), `test:mcp` 3156/0-fail, `test:prompts` 202/202, `check:syntax` clean, **live lab fire still mints a signed MEDIUM** (MAC-valid, no identity PII), 10× flake loop 0/10.

@codex review
@coderabbitai review

vmihalis:
@coderabbitai review

vmihalis:
Applied in `547eb25`.

**Note on the review path:** CodeRabbit's GitHub App stalled on `570b50e` (~1.5 hr, two explicit re-triggers, no rate-limit message), so I ran the same CodeRabbit engine via the `cr` CLI scoped to the round-16 increment (`--base-commit 6343bc4 --type committed`). It returned **exactly one finding** — a test-hardening quick win: the credential-named `canary_field` sub-case asserted the reject reason but not the no-write invariant. Applied CodeRabbit's exact suggestion (`assert.equal(mock.postCount(), 0)`). Test-only; the producer code is unchanged from `570b50e` (which Codex + the brutalist CI both cleared with zero findings).

Re-verifying clean on the increment now.

@codex review
@coderabbitai review

PR #163, opened 2026-06-25
ci: add GLM routed critic

ejmockler:
Adds a **GLM** routed Claude critic (4th critic) via the public Anthropic-compatible gateway (`immersivecommons13.tail5da903.ts.net`, a public Tailscale Funnel). Additive + hardened by default (no web/MCP, no native creds).

**Validated end-to-end on the dogfood** (ejmockler/brutalist-mcp #16): GLM authenticates, routes, stays hardened (confirmed from inside CI), and produces real findings. The `GLM_TOKEN` secret is already set here.

Caveat: GLM is the **slowest critic** (~5 min even healthy); reviews are advisory, so it never gates merges.

## Summary by CodeRabbit

* **Chores**
  * Updated automation settings to support an additional review configuration.

### DIFF
--- a/.github/workflows/brutalist-review.yml
+++ b/.github/workflows/brutalist-review.yml
@@ -82,6 +82,16 @@ jobs:
           codex-auth: ${{ secrets.CODEX_AUTH }}
           agy-oauth-token: ${{ secrets.AGY_OAUTH_TOKEN }}
           minimum-severity: medium
+          # GLM routed critic — Anthropic-compatible gateway (public Tailscale Funnel).
+          # Additive 4th critic, hardened by default. GLM is the slowest critic
+          # (~5min healthy); down=fast preflight, over-quota 429 retries ~3-4min/chunk.
+          # Reviews are advisory, so this never gates merges.
+          custom-claude-client-id: glm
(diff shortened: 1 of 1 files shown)


ejmockler:
Dispositioning the GLM review findings before merge.

GLM is **validated working across the fleet**: it authenticates, routes, stays **hardened** (verified from inside CI — no web/MCP, no native creds), and produces real findings.

**Accepted / by-design** (acknowledged, not blocking):
- **Single public-funnel gateway as a dependency** — accepted owner-risk: GLM is an *additive, advisory* critic. If it's down the native critics are unaffected and reviews don't gate merges. The diff goes to a Claude-routed client exactly like the native Claude critic.
- **GLM is the slowest critic (~5 min)** — known tradeoff for the extra perspective; advisory, so it never gates merges. (A per-client timeout would cut off legitimately-slow *healthy* runs, so not applied.)
- **Governing-window shrink to 128k** — intentional: the chunker sizes to the smallest participant window so every critic reviews every chunk.
- **Missing-token / model-fallback** — the `GLM_TOKEN` secret is set; `glm-5.1` is verified against the gateway. No fallback is inherent to an additive critic.
- The latency comment was already corrected (down = fast preflight vs over-quota 429 retry).

Resolving the threads accordingly.

PR #164, opened 2026-06-25
ci: bump brutalist-mcp to 1.18.3 (routed fail-fast)

ejmockler:
Bumps `@brutalist/mcp` to **1.18.3**: an over-quota routed critic (GLM) now **fails fast** at pre-flight (a tiny `/v1/messages` 429 probe) instead of retrying the 429 for ~3–4 min per chunk. A healthy GLM is unaffected (probe returns 200 → it runs).

### DIFF
--- a/.github/brutalist-tools/package-lock.json
+++ b/.github/brutalist-tools/package-lock.json
@@ -9,7 +9,7 @@
       "version": "0.0.0",
       "dependencies": {
         "@anthropic-ai/claude-code": "2.1.177",
-        "@brutalist/mcp": "1.18.2",
+        "@brutalist/mcp": "1.18.3",
         "@openai/codex": "0.139.0"
       }
     },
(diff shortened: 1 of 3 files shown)


ejmockler:
Re-pin to **1.18.3** — over-quota routed-critic fail-fast.

**Validated live across this rollout:** a *healthy* GLM still runs (dogfood ✅, commons ✅ — no false-skip), and an *over-quota* GLM now **fails fast at pre-flight** (bobnetsec: *"pre-flight quota check returned 429"*, **0ms** — no 3–4 min retry). (hacker-bob's GLM hit the 10-min cap only because this PR's diff is the large *regenerated lockfile*; normal PR diffs are faster.)

Findings are accepted / by-design — the routed-critic latency + single-gateway-dependency tradeoffs were already dispositioned in the GLM rollout. Resolving threads accordingly.

PR #168, opened 2026-06-26
fix(telemetry): dedupe phantom missing_marker rows per subagent (decouple from exit(2))

vmihalis:
## The bug

The `SubagentStop` hook (`.claude/hooks/agent-run-stop.js`) ends in `block()`:

```js
function block(reason, telemetryInput = null) {
  recordAgentCompletionTelemetry(telemetryInput);
  console.error(reason);
  process.exit(2);
}
```

The `exit(2)` is a **contract enforcer** — Claude Code re-invokes the stopped
subagent, which stops again, which re-fires `block()`. Every fire emits an
**identical null-coord `missing_marker`** telemetry row (no marker ⇒ no
`target_domain`/`wave`/`agent`). In one real run this produced **3431 rows that
collapsed to only 37 distinct subagents** (storms up to 1208 rows for a single
subagent). The telemetry "lies": ~90% phantom failure rate when the true
failure rate is ~0.

## The fix (decoupled from `exit(2)`)

`.claude/hooks/agent-run-stop.js` is **byte-for-byte untouched** — it is not in
this diff. `block()` and its `process.exit(2)` are provably preserved. The fix
lives entirely in the lib choke point both the hook and `bob_finalize_agent_run`
already funnel through.

**`mcp/lib/agent-run-completion.js`** — in `recordAgentCompletionTelemetry()`,
between building the telemetry `input` and the two sink calls, a gate:

```js
if (isPhantomNoMarkerRow(input) && !recordFirstPhantomMissingMarker(input)) {
  return input; // suppressed duplicate; counter already incremented
}
```

- **`transcript_path` is the discriminator.** The first stop per subagent
  (keyed by a sha256 of `transcript_path`) is recorded; re-fires are suppressed
  and counted in a small JSON counter file.
- **Only the phantom shape is gated**: `status === "blocked"` AND
  `blockCode === "missing_marker"` AND `target_domain == null && wave == null &&
  agent == null`. A real-coord blocked row (e.g. `missing_handoff` with a
  `target_domain`/`wave`/`agent`) is never gated.

**`mcp/lib/paths.js`** — `agentRunStopSuppressedPath()`, a single source of
truth for the counter path under the telemetry dir (these rows have no
`target_domain`, so there is no session dir to key them to).

**`mcp/lib/tool-telemetry.js`** — surfaces `suppressed_phantom_missing_marker`
under `bob_read_tool_telemetry` (fail-safe to 0) so a near-zero blocked rate
can't mask a storm of suppressed re-fires.

## Why it cannot hide a real failure

1. **First occurrence is always recorded** — the very first stop per subagent
   writes its row; only verbatim re-fires of the *same* subagent are dropped.
2. **Per-subagent discriminator** — distinct `transcript_path`s never
   cross-suppress (T2 proves this).
3. **Fail-OPEN** — any FS/telemetry fault in the dedupe helper returns `true`
   (record the row). A bug in this fix can only *over-record*, never suppress.
   `readJsonFile` has no default-on-missing mode (it stat-throws on an absent
   file), so the first-occurrence case is handled explicitly via `fs.existsSync`
   so it both records the row AND seeds the counter for later re-fires.
4. **Suppressed re-fires are counted and observable** via
   `bob_read_tool_telemetry`.
5. **`exit(2)` is untouched** — it lives in `block()`; the gate only decides
   whether the telemetry *row* is written.

## Tests (`test/mcp-server.test.js`)

- **T1 — storm dedupe + counter + observability**: two stops in one temp home
  with the *same* `transcript_path` and no final marker ⇒ both exit status 2
  (contract preserved both times), the tool-invocation JSONL has exactly **1**
  row, the counter file `agent-run-stop-suppressed.json` has
  `suppressed_total === 1`, and the count surfaces as
  `agent_runs.suppressed_phantom_missing_marker === 1` via
  `bob_read_tool_telemetry`.
- **T2 — over-suppression guard**: two stops with *different* `transcript_path`
  ⇒ **2** rows, `suppressed_total === 0`.
- **T3 — lib unit (no spawn)**: two phantom-shaped inputs sharing a
  `transcript_path` ⇒ 1 row + `suppressed_total === 1`; then a real-coord
  blocked input (`missing_handoff` WITH `target_domain`/`wave`/`agent`) is NOT
  gated (always recorded), and does not touch the suppression counter.

Existing regressions stay green, including **"blocks missing structured
handoff"** (real coords ⇒ still emits its row) and the **unwritable-telemetry-dir
"No marker"** test (proves `exit(2)` ⟂ telemetry).

## Verification

- `npm run check:syntax` ✅
- `npm run test:mcp` ✅ — 3168 pass, 0 fail, 2 pre-existing skips
- `npm run test:prompts` ✅ — 202 pass (write-guard manifest stays byte-identical
  to the `paths.js` projection)

## Summary by CodeRabbit

* **New Features**
  * Added smarter handling for repeated blocked completion events, reducing duplicate telemetry while keeping the original stop behavior intact.
  * Added a new lifetime telemetry metric that shows how many duplicate blocked rows were suppressed.

* **Bug Fixes**
  * Prevented repeated “phantom” events from being counted multiple times for the same transcript.
  * Added fail-safe behavior so telemetry still records when deduping data can’t be read or written.


vmihalis:
@codex review

vmihalis:
@coderabbitai review

vmihalis:
## Review revision — `bb0e7d1`

Reworked the dedupe from the single shared-JSON counter to **per-transcript marker files** under `agent-run-stop-seen/<sha256(transcript_path)[:16]>`. Core insight: storm re-fires within one transcript are **sequential** (stop → hook → `exit(2)` → continue → stop …), and concurrency only happens **across** transcripts (distinct `transcript_path` → distinct marker file). So no cross-process lock is needed — distinct subagents never share a mutable structure. `.claude/hooks/agent-run-stop.js` (`block()` / `process.exit(2)`) is **byte-for-byte untouched**; this only decides whether the telemetry ROW is written.

### Finding → fix

| # | Finding (severity / reviewer) | Fix |
|---|---|---|
| 1 | **HIGH** concurrency race / lost updates (agy + Codex + CodeRabbit Major) — shared-JSON RMW clobbers concurrent writes, can drop a `transcripts` key | Removed the shared JSON entirely. One marker file per **distinct** transcript; concurrent hook processes hit **different** files. Same-transcript re-fires (the only writers to one marker) are sequential, so the per-marker count RMW has no concurrent writer. No shared mutable counter anywhere. |
| 2 | **P2** fail-open ordering (Codex) — key seeded BEFORE the row is written; sink failure suppresses the real first row | **Row-before-mark.** `recordRow()` now returns `true` iff the invocation row was actually written; the marker is created only `if (recordRow() !== true) return;` — i.e. **after** a written first row. A sink failure leaves the transcript unmarked, so the next re-fire records again. |
| 3 | **P2/Minor** keyless collision (Codex/CodeRabbit) — empty/missing `transcript_path` hashes `""`, collapsing all such rows onto one key | `phantomTranscriptKey()` returns `null` for missing/non-string/blank `transcript_path`; the gate then **fails open** (records, never dedupes). Keyless rows can never cross-suppress. New test asserts two keyless stops → 2 rows, no marker dir. |
| 4 | **P2** telemetry opt-out (Codex) — dedupe wrote counter/marker even with telemetry disabled | `recordPhantomBlockRowOnce()` checks `telemetryEnabled(env)` first and writes **nothing** when disabled; the row sink honors the same opt-out. New test: `BOUNTY_TELEMETRY=0` → no row, no marker. |
| 5 | **medium** predicate too narrow (Claude) — null-coord `malformed_marker` storms escape the `missing_marker`-only predicate | Widened `isPhantomNullCoordBlockRow` to `missing_marker` **or** `malformed_marker`, gated on **all-coords-null**. Verified in-code that a real-coord `malformed_marker` (e.g. a `w0` wave failing the positive-`wN` regex) keeps its coordinates and is **never** gated — new test covers both. |
| 6 | **medium** aggregation-scope mismatch (Claude) — lifetime total presented as a peer of filter-scoped counts | Renamed to `suppressed_phantom_block_rows: { scope: "lifetime_global", total }` so it can't be misread as a peer of the filter-scoped `total_runs` / `by_status` / `by_block_code`. |
| 7 | **medium** unbounded growth (agy) | One small marker file per **distinct** phantom subagent (not per re-fire); re-fires only bump that file's own counter. Lives in the telemetry dir and is operator-cleaned like the rest of it (no rotation/cron added). |

### Mechanism (fail-open + race-freedom in 3 sentences)
First occurrence (marker absent): write the row via the normal sinks **first**, then create the marker — so a marker exists only after a row was actually written (closes the fail-closed hole). Duplicate (marker present): suppress the row and bump that transcript's own counter via a sequential RMW (no concurrent writer for the same key). The lifetime suppressed total is derived at **read** time in `tool-telemetry.js` by summing the per-marker counts (no shared mutable counter), fail-safe to 0 on any read error.

### Verification
- `npm run check:syntax` ✓
- `npm run test:mcp` ✓ (3172 pass / 0 fail — incl. regressions: `blocks missing final marker`, `rejects malformed, zero wave, and zero agent markers`, `blocks missing structured handoff`, and `telemetry can be disabled and write failures do not alter hook results`)
- `npm run test:prompts` ✓ (202 pass — paths.js write-guard manifest projection still byte-identical)

Tests updated T1–T3 to the new mechanism and added: fail-open ordering (broken first-row sink → not marked → repaired second stop records), keyless skip, telemetry opt-out, and widened-predicate (null-coord `malformed_marker` dedupes; real-coord never gated).

vmihalis:
@codex review

vmihalis:
@coderabbitai review

vmihalis:
Latest push hardens the observability reader; the rest of the review was already addressed by the redesign in `bb0e7d1`. Resolving all threads:

**New this round (fixed in the latest commit):**
- **[Claude 🟡 medium] perf — `readLifetimeSuppressedPhantomBlockRows` is unbounded O(N) on every summary** and **[Codex 🟡 medium] perf — no filename filter / entry cap / size / symlink validation** → reader now filters to the marker format `^[0-9a-f]{16}$`, caps the scan, `lstatSync`-skips symlinks/non-regular files (never follows a symlink), and size-bounds each marker before parse. Still fail-safe to 0.

**Already addressed by the `bb0e7d1` redesign (these threads predate it):**
- **[agy/Codex 🟠 HIGH] read-modify-write race / lost-update counter** → replaced the shared JSON counter with **per-transcript marker files**; storm re-fires within one transcript are sequential and distinct transcripts hash to distinct files, so there is no shared mutable state to race on, and the lifetime total is derived by summing per-marker counters at read time.
- **[Codex P2] do not seed dedupe before the row is written** → **row-before-mark**: the marker is created only after `recordRow()` confirms the invocation row was actually written (`=== true`), so a sink failure leaves the transcript unmarked and the next re-fire records again.
- **[Codex/CodeRabbit] keyless empty-string collision** → missing/blank `transcript_path` now fails OPEN (records, never deduped).
- **[Codex P2] honor `BOUNTY_TELEMETRY=0`** → `telemetryEnabled(env)` gate writes nothing when telemetry is off.
- **[Claude 🟡] predicate too narrow** → widened to null-coord `malformed_marker` (real-coord malformed keeps its coords and is never gated).
- **[Claude 🟡] aggregation-scope mismatch** → surfaced as `suppressed_phantom_block_rows: { scope: "lifetime_global", total }`.
- **[agy 🟡] unbounded growth** → one small file per DISTINCT phantom subagent, operator-cleaned like the rest of the telemetry dir.

`.claude/hooks/agent-run-stop.js` (`block()`/`exit(2)`) remains untouched. `test:mcp` 3172 pass / 0 fail. @codex review @coderabbitai review

ejmockler:
Ported the transcript-completion deduplication to the current mcp/core runtime, retained fanout/evidence telemetry behavior, added regression coverage, and landed it on private main at e944cc13; public synchronization is #186.

PR #166, opened 2026-06-25
fix(evaluator): route fail-closed offensive producers on surface signal

vmihalis:
## Problem

The first live offensive-measurement run (`/bob-evaluate https://www.dromeas.gr --deep`, 2026-06-25) invoked the signed offensive producers **0 times across 267 `bob_http_scan` calls** — and not just the auth-gated ones: even the **auth-free, fail-closed differential producers** (`bob_http_xss_reflect` / `bob_http_cors_confirm`) never fired. No `offensive-runs.jsonl` was written. The run still graded SUBMIT 85, but entirely via the classic read-only path — the offensive arsenal it was launched to exercise got zero live invocations (same outcome as the prior vu.nl run).

**Root cause (forensic post-mortem, adversarially verified):** the evaluator role framed producers as a *confirm-once-you-already-believe* step — *"when you have proven or strongly suspect a primitive … confirm it with the matching producer."* So evaluators hand-rolled prose `bob_http_scan` differentials and never routed the suspected signal through the signed adjudicators. The binding gate was framing, not transport or identity (the fail-closed producers need neither).

## Change

Reframe the **fail-closed differential producers** as the PRIMARY adjudication path, **routed on surface signal** rather than on prior belief in a mint:
- reflected query params → run `bob_http_xss_reflect`
- a response reflecting the request `Origin` in `Access-Control-Allow-Origin` → run `bob_http_cors_confirm`
- run them **in addition to** any manual `bob_http_scan` differential — the producer is what mints the signed, tamper-evident proof, so a hand-rolled scan never substitutes for it.

The **auth-gated / out-of-band** producers are explicitly kept under stronger gating (NOT free to fire blindly): the OOB pair (self-hit caveat), `bob_http_massread_confirm` (needs a provisioned under-privileged identity), and `bob_http_idor_confirm` (operator arming + synthetic identities). Those still gate on a concrete hypothesis + their stated prerequisites.

## Safety

Routing the fail-closed producers more often **cannot mint on a non-vulnerable surface** — this PR does not touch any mint condition; it only changes *when the evaluator calls the tool*. A non-mint stays "not proven here", never "safe". The server-side invariants (per-row HMAC, server-stamped `demonstrated_severity` ceiling, the `#18` synthetic-provenance gate, the IDOR arming env) are untouched.

## Mechanics

- Source edit: `prompts/roles/evaluator.md` + `prompts/roles/evaluator-spawn.md`.
[...]

PR #169, opened 2026-06-26
fix(offensive): helpful surface_id resolution error so producers self-correct

vmihalis:
## What

`findRoutedSurface` (shared by all signed producers — reflect/cors/massread/idor) threw a bare `unknown or unrouted surface_id X` with **no list of valid ids and no suggestion**.

## Why it matters (found by the 2026-06-26 routing smoke-test)

A headless evaluator carrying the merged-#166 prompt **correctly routed** to `bob_http_xss_reflect` / `bob_http_cors_confirm` on a reflected-param / CORS signal — but passed `surface_id:"search"` instead of `"surface:search"`. The producer errored, the error gave no hint, and the agent **abandoned the producer and fell back to a hand-rolled scan + manual claim** — the exact under-firing the signed arsenal is meant to replace. Routing worked; firing was lost on a recoverable input slip.

## The fix

The error now:
- lists the routed `surface_ids` (capped at 20, with a `(N total)` suffix),
- suggests the closest match (`did you mean surface:search?`) — matching a dropped `surface:` prefix first,
- carries `routed_surface_ids` + `suggested_surface_id` in `details`.

So an agent that slips on the id format can **self-correct and fire** instead of giving up.

## Tests

- `test/offensive-http-common.test.js`: strengthened the existing unrouted-id test (asserts the routed-id list + `details.routed_surface_ids`); added a regression test for the dropped-prefix suggestion.
- `npm run test:mcp` green (3168 tests, 0 fail); `check:syntax` clean.

Behavior change is error-message-only; no change to which ids resolve.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved error messages for unrouted surface lookups, including exact-match “did you mean …?” suggestions and guidance to re-run routing when appropriate.
  * Bounded and clarified routed `surface_id` excerpts (cap on listed items with total count), and safely clipped oversized IDs.
  * Hardened error output by sanitizing echoed `surface_id`/reason text to remove control characters and unsafe Unicode.
  * Added a dedicated error code for unrouted surface IDs (`surface_id_unrouted`) to make error handling more precise.

### DIFF
--- a/mcp/lib/offensive-http-common.js
+++ b/mcp/lib/offensive-http-common.js
@@ -369,43 +369,52 @@ function findRoutedSurface(domain, surfaceId) {
     // signed producer — falling back to a hand-rolled scan + manual claim, the exact under-firing
     // the offensive arsenal is meant to replace. The MESSAGE is the load-bearing self-correction
     // interface (the caller is an LLM reading the error text); no consumer reads a structured
-    // route inventory, so details stay at the stable {code}. Everything echoed is length-bounded
-    // so a single miss can't bloat the payload (this error deliberately drives a retry loop).
+    // route inventory, so details stay at the stable {code}. EVERY id echoed into the message —
+    // the caller's surface_id, the listed routes, the suggestion, and the quarantine hint — is
+    // passed through safe() (bound to MAX_ID_LEN AND control-char-stripped), so no echo path can
+    // bloat the line or forge extra message/log content. Uniform: no echo path is exempt.
     const MAX_LISTED = 20;
     const MAX_ID_LEN = 120;
-    const clip = (id) => (id.length > MAX_ID_LEN ? `${id.slice(0, MAX_ID_LEN - 1)}…` : id);
-    // The caller's surface_id is untrusted free-text: bound it AND strip control chars, so a huge
-    // or newline-laden value can neither bloat the line nor forge extra message/log content.
-    const safeId = clip(surfaceId).replace(/[\x00-\x1f\x7f]/g, "·");
+    const safe = (id) => {
+      const clipped = id.length > MAX_ID_LEN ? `${id.slice(0, MAX_ID_LEN - 1)}…` : id;
+      return clipped.replace(/[\x00-\x1f\x7f]/g, "·");
+    };
     const known = routed.document.routes
       .map((entry) => entry.surface_id)
       .filter((id) => typeof id === "string" && id);
-    // Suggest the closest routed id by EXACT EQUALITY only (the dropped `surface:` prefix or its
-    // inverse) — never a substring/suffix wildcard, which for a short/typo'd input could point at
-    // the wrong (if in-scope) surface, against this file's {id}-must-terminate discipline.
-    const suggestion = known.find((id) => id === `surface:${surfaceId}`)
-      || known.find((id) => id === surfaceId.replace(/^surface:/, ""))
-      || null;
-    // If no LIVE route matches but a QUARANTINED (malformed) route matches the same exact
-    // relation, the surface exists — it just needs re-routing; say so rather than a bare
-    // "unknown" (the exact-id malformed branch above only covers the canonical-id caller; a
-    // surface that still has a valid route resolves above and never reaches here).
-    const quarantinedMatch = suggestion
-      ? null
-      : (Array.isArray(routed.malformed_routes) ? routed.malformed_routes : [])
-        .map((entry) => entry && entry.surface_id)
-        .filter((id) => typeof id === "string" && id)
-        .find((id) => id === `surface:${surfaceId}` || id === surfaceId.replace(/^surface:/, "")) || null;
+    const quarantined = (Array.isArray(routed.malformed_routes) ? routed.malformed_routes : [])
+      .map((entry) => entry && entry.surface_id)
+      .filter((id) => typeof id === "string" && id);
+    // Suggest the closest id by EXACT EQUALITY only (the dropped `surface:` prefix or its inverse)
+    // — never a substring/suffix wildcard, which for a short/typo'd input could point at the wrong
+    // (if in-scope) surface, against this file's {id}-must-terminate discipline. An exact match is
+    // only possible within ±"surface:".length of a candidate id's length, so for anything longer we
+    // skip the key-building entirely — it cannot match, and this bounds both the work and the
+    // rebuilt `surface:${surfaceId}` string on a pathologically long surface_id.
+    const maxCandidateLen = [...known, ...quarantined].reduce((max, id) => Math.max(max, id.length), 0);
+    const canMatch = surfaceId.length <= maxCandidateLen + "surface:".length;
+    const matchExact = (ids) => (canMatch
+      ? ids.find((id) => id === `surface:${surfaceId}`)
+        || ids.find((id) => id === surfaceId.replace(/^surface:/, ""))
+        || null
+      : null);
+    // Prefer a LIVE routed id (the caller can act on it immediately). Only when no live id matches
+    // do we point at a QUARANTINED (malformed) duplicate — a surface that exists but needs
+    // re-routing. A surface that still has a valid route resolves above and never reaches here, so a
+    // live suggestion is always the most actionable hint even when a malformed duplicate of the
+    // same id also exists.
+    const suggestion = matchExact(known);
+    const quarantinedMatch = suggestion ? null : matchExact(quarantined);
     const listed = known.length
-      ? known.slice(0, MAX_LISTED).map(clip).join(", ") + (known.length > MAX_LISTED ? `, … (${known.length} total)` : "")
+      ? known.slice(0, MAX_LISTED).map(safe).join(", ") + (known.length > MAX_LISTED ? `, … (${known.length} total)` : "")
       : "(none)";
     const hint = suggestion
-      ? ` (did you mean ${clip(suggestion)}?)`
+      ? ` (did you mean ${safe(suggestion)}?)`
       : quarantinedMatch
-        ? ` (${clip(quarantinedMatch)} exists but has a malformed route — re-run bob_route_surfaces)`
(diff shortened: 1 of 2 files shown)


vmihalis:
@codex review
@coderabbitai review

vmihalis:
Addressed both brutalist findings in `a6c8ab2`:

1. **Bare-substring `includes` suggestion → dropped.** It could point a short/typo'd `surface_id` ("a", "api") at the wrong (if in-scope) surface, against the file's `{id}`-must-terminate discipline. The exact-prefix / `:suffix` / de-prefix matchers still cover the real slip (a dropped `surface:` prefix).
2. **Unbounded `details.routed_surface_ids` → capped.** `envelope.js` reflects `details` verbatim with no size bound, so the message-only `MAX_LISTED` cap still dumped the whole route table on every miss (amplification footgun). Now `routed_surface_ids: known.slice(0, MAX_LISTED)` + `routed_surface_count: known.length`.

Kept `details.code` (matches the existing `{code: ...}` details convention elsewhere in the producers; the nit is acknowledged-harmless since `classifyException` keys off the top-level `error.code` which stays `INVALID_ARGUMENTS`). `test:mcp` green.

@codex review
@coderabbitai review

vmihalis:
Addressed both r2 brutalist findings in `8f47099`:

1. **Suffix-wildcard suggestion → removed.** `suggested_surface_id` is now EXACT EQUALITY only (the dropped `surface:` prefix or its inverse). No `endsWith()` arm, so it can't point at a different-but-suffix-sharing surface. Comment corrected to match.
2. **Byte-bound the list.** Added `MAX_ID_LEN=120` per-id clip on the emitted list (message + `details.routed_surface_ids`), so a single oversized `surface_id` can't amplify the payload — not just count-capped now. The actionable `suggested_surface_id` is left whole for verbatim copy.

`test:mcp`/`check:syntax` green.

@codex review
@coderabbitai review

vmihalis:
Addressed both r3 brutalist findings in `964352d`:

1. **Quarantined-AND-dropped-prefix self-correction gap → fixed.** When no live route matches but a QUARANTINED (malformed) route matches the same exact relation, the error now says `<id> exists but has a malformed route — re-run bob_route_surfaces` (+ `details.quarantined_match`). So a dropped-prefix caller against a corrupt route file gets actionable guidance, not a bare "unknown". (The exact-id malformed branch above only covered the canonical-id caller.)
2. **Untested numeric bounds → covered.** Added 3 tests: count cap (20 emitted + full count + `(N total)`), per-id length clip (120 + ellipsis), and the quarantine hint.

`test:mcp` green (3169/0).

@codex review
@coderabbitai review

vmihalis:
Addressed the r4 findings in `01424c9` — a simplification, per the design call:

1. **Write-only structured details → dropped.** Correct observation: the caller is an LLM reading the error *text*, and nothing consumed `routed_surface_ids`/`suggested_surface_id`/`quarantined_match`. Details are now just the stable `{code}`. This also removes the byte-cap surface those arrays created.
2. **Unbounded/unsanitized `${surfaceId}` echo → fixed.** The caller id is now clipped to 120 chars AND has control chars stripped (`/[\x00-\x1f\x7f]/` → `·`), so a huge or newline-laden value can't bloat or forge the line. The suggestion/quarantine ids in the message are clipped too — everything echoed is bounded.
3. **"Check quarantined before live" — respectfully, the current order is correct.** The live suggestion is checked *first* by design: when a valid route exists for the canonical id it IS the answer (it resolves directly, or is suggested), and the quarantined duplicate is only surfaced when NO live route exists. Checking quarantine first would wrongly steer a caller to re-route when a working route already exists. So this is intended, not a defect.

Tests now assert on the message (the real interface) + a new bound-and-sanitize test. `test:mcp` green.

@codex review
@coderabbitai review

vmihalis:
## Round-5 findings → addressed in aaad3f1

The round-5 threads (Codex P2 ×3 + brutalist medium ×3) converged on one real gap — the round-4 self-correction error sanitized only the **caller's** surface_id, not the other echoed ids — plus a lookup-bound and a test-rigor point.

| Finding | Resolution (aaad3f1) |
|---|---|
| Codex — "Sanitize listed route ids before echoing them" | **Fixed.** One `safe()` (clip to MAX_ID_LEN **and** strip control chars) now guards **every** echoed id — caller, listed, suggestion, quarantine. |
| brutalist (security) — "Control-char sanitization is asymmetric: only the caller id is stripped" | **Fixed** — same uniform `safe()`. |
| brutalist (security) — "'No forging' holds for 1 of 4 echo paths; the comment overstates" | **Fixed** — uniform `safe()` + the comment rewritten to state all four paths are sanitized (no overstatement). |
| Codex — "Bound the lookup key before route suggestions" | **Fixed.** An exact match is only possible within ±`"surface:"` of a candidate id's length, so a longer caller id skips key-building entirely — no `surface:${id}`/`replace` on unbounded input. New test: a 100k-char id → no suggestion, clipped echo, message < 1 KB. |
| brutalist (testing) — "Tests are shaped to pass: the sanitize test asserts a property it structurally cannot verify" | **Fixed.** Tests now exercise a **real** control char (BEL) on every **non-caller** echo path: listed + suggestion via a BEL-bearing routed id (probed: it survives routing into the live inventory), and the quarantine hint via a hand-written malformed route. The property is demonstrated, not vacuous. |
| Codex — "Check quarantined duplicates before suggesting live ids" | **Working as intended — re-justified.** A surface with a valid route resolves at the **top** of `findRoutedSurface` and never reaches this branch; reaching here with a live exact-match means the canonical id **is** usable, so suggesting it lets the caller immediately re-fire. The quarantine hint is the fallback for when **no** live route matches. Surfacing "re-run routing" when a valid route already exists would be misdirection. A clarifying comment now states the live-first ordering is intentional. |

Verified: `node --test test/offensive-http-common.test.js` → **30/30**; full `npm run test:mcp` → **3174 tests, 0 fail**; no raw control bytes in source; rebased on `main` so the package-size gate passes.

Resolving the six threads.

vmihalis:
@codex review
@coderabbitai review

vmihalis:
## Round-6 finding → addressed in 058f9e2

**brutalist 🟠 HIGH (correctness): "The hardening is an island — the 'Uniform' invariant is falsified two scopes up."** Correct, and a good catch. `findRoutedSurface` has **three** surface_id echo sites and round 5 hardened only the middle one:

- malformed-route rejection (line 362) — echoed **raw** surfaceId + **raw** `malformed.reason`
- unrouted rejection — hardened in r5 ✓
- routed-but-absent rejection (line 424) — echoed **raw** surfaceId

So "Uniform: no echo path is exempt" was a load-bearing falsehood. Fix:

- **Hoisted `safe()` above the first rejection** (now `String()`-coercing so it also covers the malformed reason) → the guarantee is genuinely function-wide. The "uniform" comment moved to the declaration, where it is now true.
- **All three rejections route through `safe()`**; `malformed.reason` is sanitized too (validator-generated, lower risk, but covered for genuine uniformity).
- Gave the two other rejections stable `{code}`s (`surface_id_malformed_route`, `surface_id_not_in_surfaces`) matching the unrouted branch's `{code}` discipline. Clean-id callers get **byte-identical** messages.
- **Tests now DEMONSTRATE the invariant on all three sites** — added malformed-route + routed-but-absent control-char sanitize tests (a BEL id that exact-matches a quarantined route → malformed branch; a BEL id routed but dropped from `attack_surface.json` → absent branch), alongside the caller/listed/suggestion/quarantine coverage.

Verified: `node --test test/offensive-http-common.test.js` → **32/32**; full `npm run test:mcp` → **3176 tests, 0 fail**; no raw control bytes in source.

Resolving the thread.

vmihalis:
@codex review
@coderabbitai review

vmihalis:
## Round-7 findings → addressed in dbe61d1

Codex P2 ("Strip Unicode line/bidi controls from echoed surface IDs") + brutalist 🟡 medium ("Bidirectional Unicode override bypass") — same gap, and legitimate: `safe()`'s `/[\x00-\x1f\x7f]/` stripped only **ASCII** controls. A surface_id carrying a Unicode **line separator** (U+2028/9) could still forge a log line, and a **bidi override** (U+202A–E / U+2066–9 / LRM·RLM·ALM — the Trojan-Source set) could visually reorder the echoed id. The hoisted comment already *claimed* "RTL-override" protection it didn't deliver — the same overstatement defect, now closed.

- Extended the strip class to `\x00-\x1f`, `\x7f-\x9f` (DEL + C1), `U+061C`, `U+200E/F`, `U+2028/9`, `U+202A-E`, `U+2066-9` — the complete invisible / forging / reordering set, so the "no forging / no visual reorder" claim is now true.
- The comment names the set, so the coverage is auditable.
- Test (built with `String.fromCharCode`, so the raw chars never enter the source) feeds an `RLO + LINE SEPARATOR + PDI` surface_id and asserts each becomes `·` and none survives raw.

Verified: `node --test test/offensive-http-common.test.js` → **33/33**; full `npm run test:mcp` → **3177 tests, 0 fail**; no raw control/bidi bytes in source.

Resolving both threads.

vmihalis:
@codex review
@coderabbitai review

vmihalis:
## Round-8 → addressed in 82fd832

**CodeRabbit 🟠 Major: "Assert the exact 120-character clip, not just 'not the full id.'"** Fair — the bounded-lookup test only proved the full id was omitted, so a drift to 121/130/any shorter-than-full bound would still pass. Now pins the EXACT clip: the echoed id includes `'z'×119 + '…'` AND **not** `'z'×120`, so any change to MAX_ID_LEN is caught.

Also note: **brutalist passed clean this round** (the sanitization axis converged). One `test (20)` run flaked on an **unrelated** OOB test (`poll IDEMPOTENT re-poll`) — passed locally (3177/0) and on the other 3 CI runs; this push re-runs it.

Verified: `node --test test/offensive-http-common.test.js` → 33/33. Resolving the thread.

vmihalis:
@coderabbitai review

PR #171, opened 2026-06-27
fix(authority): thread lab_authorization through the bootstrap gate (init deadlock)

vmihalis:
## Why

The **first full orchestrated lab run** (`/bob-evaluate` against a local `127.0.0.1:8899` lab) exposed a real init bug that none of the direct-function smoke-tests could: `bob_init_session` for an operator-attested private lab **hard-deadlocked**.

Root cause: `authorizeBootstrap` — the pre-handler MCP authority gate — runs **before** the session handler and called `assertHttpScopeDomain` / `validateHttpScanScope` **without** `lab_authorization`. The handler threads it; the gate didn't. So a fresh private-lab init was rejected at the gate, and the persisted `lab-authorization.json` that the fallback path reads is only ever written by the very handler the gate was blocking — a chicken-and-egg deadlock. The smoke harnesses call `initSession()` directly and bypass this gate, so only a real MCP dispatch reaches it.

## What

Thread the **same fail-closed** `parseLabAuthorization(args.lab_authorization)` the handler already uses through the bootstrap gate (`normalizeArgumentTarget` → `assertHttpScopeDomain`, and `validateHttpScanScope`).

`parseLabAuthorization` reads the operator env ack (`BOB_LAB_TARGET_ACK`) **first**, before touching any arg, and returns `null` unless ack + `private_targets:true` + host-match (`BOB_LAB_TARGET`) all hold. So:
- **non-lab / public target** → `labAuthorization = null` → the gate behaves **exactly as before** (unchanged).
- **lab target** → allowed only under the full **env-ack + host-bind + in-call-intent** triad — the security model is fully intact.

## Verification

- New regression test (`test/repo-target-binding.test.js`) pins the exact hole: the bootstrap gate **ALLOWS** a `127.0.0.1` lab init with the full attestation, and still **BLOCKS** (a) without the in-call intent, (b) with a different bound host, (c) without the env ack; a public target is unaffected.
- Full `npm run test:mcp` → **3178 tests, 0 fail**.
- The fix was already **live-validated** in the run session (5/5 focused harness + 32/32 lifecycle) before being captured here for proper review.

Security-sensitive (init authority gate) — please review the fail-closed threading carefully.

## Summary by CodeRabbit

* **Bug Fixes**
  * Made bootstrap session authorization lab/private-target aware by parsing operator-provided lab authorization and using it during target normalization and HTTP scope validation.
  * Centralized attested-lab bootstrap policy checks to fail closed on missing/mismatched attestation and reject conflicting settings (e.g., internal-host blocking and non-default egress).
* **Tests**
  * Added regression coverage ensuring the bootstrap gate correctly requires/passes `lab_authorization` and rejects `target_domain`/`target_url` host drift, including additional invalid policy combinations.

### DIFF
--- a/mcp/lib/lab-target-attest.js
+++ b/mcp/lib/lab-target-attest.js
@@ -32,6 +32,9 @@
 // cannot self-grant a private-target scan by forging the file.
 
 const net = require("net");
+// Used so the lab-bootstrap validator normalizes egress_profile IDENTICALLY to the initSession
+// handler (validation.js is a leaf utility — no import cycle back to this module).
+const { assertNonEmptyString } = require("./validation.js");
 
 // Operator attestation channel. The ack is an ENVIRONMENT variable, NOT a tool
 // argument: a model that can call bob_init_session must not be able to authorize
@@ -162,15 +165,25 @@ function labBootstrapPolicyViolation(args, labAuthorization) {
     };
   }
   // (2) A proxy-backed egress would scan the attested private target from the proxy's network, not
-  // the operator's lab network — require the default (direct) egress profile.
-  const egress = args && typeof args.egress_profile === "string" && args.egress_profile.trim()
-    ? args.egress_profile.trim()
-    : "default";
+  // the operator's lab network — require the default (direct) egress profile. Normalize EXACTLY as
+  // the initSession handler does (null/undefined → "default", else assertNonEmptyString which trims
+  // and throws on a non-string/empty/whitespace value), catching that throw as a violation — so the
+  // gate rejects precisely what the handler rejects, with no "gate allowed → handler throws" gap on a
+  // malformed egress_profile (this is what makes gate-validation ⊆ handler-validation actually hold).
+  const egressViolation = {
+    code: "lab_authorization_requires_default_egress",
+    message: "lab_authorization requires the default (direct) egress profile: a proxy-backed egress would scan the attested private target from the proxy's network, not the operator's lab network",
+  };
+  let egress;
+  try {
+    egress = args == null || args.egress_profile == null
+      ? "default"
+      : assertNonEmptyString(args.egress_profile, "egress_profile");
+  } catch {
+    return egressViolation;
+  }
   if (egress !== "default") {
-    return {
-      code: "lab_authorization_requires_default_egress",
-      message: "lab_authorization requires the default (direct) egress profile: a proxy-backed egress would scan the attested private target from the proxy's network, not the operator's lab network",
-    };
+    return egressViolation;
   }
   return null;
 }
--- a/mcp/lib/lab-target-attest.js
+++ b/mcp/lib/lab-target-attest.js
@@ -146,6 +146,35 @@ function labTargetPermitted(host, { authorization } = {}) {
   return operatorAuthorizedLabHost(host);
 }
 
+// The two POLICY constraints an operator-attested lab init must satisfy, factored into ONE place so
+// the pre-handler authority gate (session-authority.js authorizeBootstrap) and the initSession
+// handler (session-state.js) enforce IDENTICAL rules — no split-brain where the gate returns
+// "allowed" for a lab init the handler then rejects. Returns a {code, message} violation descriptor
+// or null; each caller throws it in its own form (the handler a ToolError, the gate a blockedDecision).
+function labBootstrapPolicyViolation(args, labAuthorization) {
+  if (!labAuthorization) return null;
+  // (1) An attested private target needs internal-host egress to be reachable, so it cannot be
+  // combined with an explicit block_internal_hosts:true.
+  if (args && args.block_internal_hosts === true) {
+    return {
+      code: "lab_authorization_block_internal_conflict",
+      message: "lab_authorization cannot be combined with block_internal_hosts: an attested private target requires internal-host egress to be reachable",
+    };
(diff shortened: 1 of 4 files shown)


vmihalis:
@codex review
@coderabbitai review

vmihalis:
## Round-1 brutalist finding → addressed in aca1adc

**brutalist 🟡 medium (testing): "Regression test opens the private-target surface but never pins the SSRF-pivot block (host != domain)."** Sharp catch — that's the exact case this fix first activates. Added it: WITH the full `127.0.0.1` attestation, a `target_url` on **169.254.169.254** (cloud-metadata) or a **non-attested RFC1918 host (10.9.9.9)** must still be **blocked**. The test **passes**, which proves the lab-eligibility relaxation does NOT open an SSRF pivot — `validateHttpScanScope`'s url-drift check is independent of it. So this both closes the coverage gap and turns the concern into a positive proof.

Resolving the thread.

vmihalis:
@codex review
@coderabbitai review

vmihalis:
## Round-2 finding → addressed in c9c9dbb

**brutalist 🟡 medium (agy + glm + Codex, all agree): "Split-brain — the gate threads lab_authorization for the SCOPE checks but drops the handler's two lab POLICY checks."** Correct, and it's the "comment claims an invariant that's now false" pattern: my fix's comment said the gate mirrors the handler, but it only mirrored the scope checks — so a lab init with `block_internal_hosts:true` or a non-default egress returned `authority_result:'allowed'` at the gate and was then thrown by the handler. (Telemetry-only today, as all three noted — but a future fast-path trusting the gate would inherit the gap.)

Fixed exactly as recommended — **one shared validator, both sites call it, no copied logic**:
- Extracted **`labBootstrapPolicyViolation(args, labAuthorization)`** into `lab-target-attest.js` (single source of the two lab policy constraints).
- The handler's two inline checks collapse to one call to it (messages **byte-identical**); the gate runs the same validator **before** scope normalization, matching the handler's error order.
- Test pins **gate ⊆ handler**: the gate now rejects `lab+block_internal_hosts` and `lab+non-default-egress`.

Verified: `lab-target-attest` + `repo-target-binding` 29/29; full `npm run test:mcp` **3178, 0 fail**.

Resolving the thread.

vmihalis:
@codex review
@coderabbitai review

vmihalis:
## Round-3 finding → addressed in 611048d

**brutalist 🟡 medium (4 critics, correctness): the shared validator's egress type-guard still diverged from the handler on MALFORMED `egress_profile`.** Verified and correct: the validator coerced any non-string/empty/whitespace value to `"default"` (→ gate allowed), while the handler's `assertNonEmptyString` throws on those — so a lab init with `egress_profile: "" / 42 / {}` passed the gate and was then rejected by the handler. (The judge confirmed the **block_internal half is FALSE** — both sites use strict `=== true`, no divergence there.)

Fixed so the two **cannot** diverge: the validator now normalizes egress via the **same `assertNonEmptyString`** the handler uses (imported from `validation.js`, a leaf util — no cycle), catching its throw as a violation. So `null/undefined` and a string trimming to `"default"` pass; every other value — non-default string, `""`, whitespace, number, object, array — is a violation, exactly mirroring the handler's reject set. It's now a genuinely shared bootstrap validator.

Test: added the malformed-egress variants (`""`, whitespace, `42`, `{}`, `["proxy"]`) under lab attestation — all blocked at the gate (the suite previously covered only the clean non-default string).

Verified: `repo-target-binding` + `lab-target-attest` 29/29; full `npm run test:mcp` **3178, 0 fail**.

Resolving the thread.

vmihalis:
@codex review
@coderabbitai review

vmihalis:
## Round-4 finding → addressed in 8441495

**brutalist 🟡 medium (glm): the gate⊆handler claim is narrower than the comment — strict `=== true` admits a non-boolean `block_internal_hosts`.** Correct (and it corrects the *prior* round's judge: `deriveBlockInternalHostsPolicy:251` **does** `assertBoolean(block_internal_hosts)`, so `block_internal_hosts:"yes"` passed the validator's `=== true` → gate allowed, then threw in the handler). Telemetry-only, but the exact overstatement to close. Did **both** fixes glm offered:

- **Mirror `assertBoolean`** in the validator (catch throw → violation), symmetric with the round-3 egress `assertNonEmptyString` mirror — so a truthy non-boolean `block_internal_hosts` is rejected at the gate too (`false`/unset still allowed).
- **Scoped the comment** to the honest guarantee: the gate agrees with the handler on the two lab POLICY checks + scope (normalizing `block_internal_hosts`/`egress_profile` identically), and is **explicitly NOT a full input validator** — other fields (`allow_internal_hosts` type, etc.) stay handler-side, reached after the gate. So the residual input-validation glm noted is now *documented as handler-side*, not a hidden divergence.

Test: lab + non-boolean `block_internal_hosts` (`"yes"`/`1`/`{}`/`"true"`) all blocked at the gate; lab + `block_internal_hosts:false` still allowed.

Verified: `repo-target-binding` + `lab-target-attest` 29/29; full `npm run test:mcp` **3178, 0 fail**. (The prior commit's red check was an unrelated OOB-poll flake, green on 3/4 runs + locally.)

Resolving the thread.

vmihalis:
@codex review
@coderabbitai review

PR #172, opened 2026-06-27
fix(offensive): IDOR producer confirms multi-endpoint single-host surfaces (live-run gap)

vmihalis:
## Why

The **first live orchestrated run** found this: `bob_http_idor_confirm` was *invoked* by the pipeline, self-provisioned its canaries, and ran the full cross-tenant differential — then **refused to sign**: the technique log shows *"signed producer non-fire on surface endpoint_count=3."* The producer was deliberately scoped to **single-ENDPOINT** surfaces ("GAP A"), but real discovery bundles a collection + item route (and more) under one surface, so it bailed before minting — even though the evaluator proved the IDOR by hand.

## What

`assertSingleEndpointSingleHost` → **`assertSingleHostBoundEndpoint(surface, stateOrigin, pathTemplate)`**:

- **Keep the single-HOST safety** — every recorded endpoint **plus** `surface.hosts` must collapse to **one origin** (unchanged: a relative endpoint + a different declared host still can't split the target across origins).
- **Allow MULTIPLE endpoints**, and bind the target to the **one** recorded endpoint that `path_template` matches — the *same* selection `resolveBaselineFromSurface` already makes via `pathTemplateMatchesEndpoint`.
- **`>1` match → `ambiguous` rejection** (refuse rather than guess). **`0` match** is left to `resolveBaselineFromSurface`'s existing AC-2 rejection right after.

`row.target` stays pinned to a **server-recorded endpoint on the surface's single declared host**, never agent free-text. No new attack surface: an agent can already only target endpoints on the one declared host — that holds whether the surface records 1 or N of them.

## Tests

- **NEW positive:** a multi-endpoint single-host surface (collection `/api/accounts` + item `/api/accounts/{id}`) now **FIRES**, with `row.target` pinned to the matched **item** endpoint (not the collection).
- **Flipped:** a surface where `path_template` matches **>1** recorded endpoint is rejected as **ambiguous** (was the old "multi-endpoint refuses" test — now it refuses for the right reason).
- Multi-**HOST** still rejected; all single-endpoint paths unchanged.
- `node --test test/offensive-idor-producer.test.js` → **170/170**; full `npm run test:mcp` → **3179, 0 fail**.

Signed producer — please scrutinize the binding (target pinning + single-host) carefully.

## Summary by CodeRabbit

* **Bug Fixes**
  * Strengthened confirmation targeting so `path_template` deterministically binds to the correct recorded item endpoint even when surfaces include multiple related endpoints.
  * Prevented unrelated or aggregated sub-resources under the same collection from being accepted.
  * Rejects unsupported endpoint URL formats and query-routed sibling endpoints to enforce clean, path-based matching.

* **Documentation**
  * Updated `bob_http_idor_confirm` guidance to allow multi-endpoint surfaces while still requiring an unambiguous `path_template`→endpoint mapping.

* **Tests**
  * Expanded acceptance and rejection scenarios, including order-independence and encoding edge cases, and verified no rows are written on failures.


vmihalis:
@codex review
@coderabbitai review

vmihalis:
## Round-1 findings → addressed in eb95ba5

| Finding | Resolution |
|---|---|
| **Codex P1 — intra-surface laundering** (relaxing to arbitrary multi-endpoint lets an accounts proof back a users finding via shared surface_id) | **Fixed.** Require every recorded endpoint to live under `path_template`'s **collection** (segment-aware, so `/api/accounts-archive` is not "under" `/api/accounts`) — the surface must represent ONE resource family. New test pins the unrelated-aggregation rejection. The collection check runs only after a unique match, so a single-endpoint off-route surface still gets the AC-2 "does not match" message. |
| **brutalist (correctness) — 0-match fallback to `endpoints[0]`** | **Fixed.** The function now rejects BOTH 0 and >1 matches, so `boundEndpoint` is always a genuine match (no coincidence-safe fallback). 0-match keeps the AC-2 `path shape does not match any recorded endpoint` phrasing. This also moots the "fallback safety depends on predicate equivalence" concern. |
| **brutalist (DRY/correctness) — local resolver diverged from `urlFromEndpoint` on `//` refs** | **Fixed the divergence:** `resolveUrl` now rejects a `//host` network-path ref (resolves to a foreign origin) exactly as `urlFromEndpoint` does; the match uses the shared `pathTemplateMatchesEndpoint`. The remaining pure resolve-loop duplication is acknowledged — collapsing it into a shared helper that returns *both* endpoint and baseline is a larger `resolveBaselineFromSurface` refactor (used by other producers), out of scope for this fix; the two now agree on the security-relevant `//` handling. |
| **Codex P2 — descriptor still said "single-endpoint only"** | **Fixed.** The MCP tool descriptor now describes the same-collection multi-endpoint relaxation, so agents actually use it. |
| brutalist (LOW design) — mixed relative + absolute-subdomain endpoints → multi-host reject | **Intentional (fail-closed).** Acknowledged as an operability cost, not a security issue, per the critic. |

Verified: `offensive-idor-producer` **171/171**; full `npm run test:mcp` **3180, 0 fail**.

Resolving the four threads.

vmihalis:
@codex review
@coderabbitai review

vmihalis:
## Round-2 findings → addressed in f4db063

All seven converged on the same root: my round-1 collection-**prefix** check was too loose, and the ground-truth lab surface (`["/api/users/:id", "/api/users/1"]`) showed my "exactly one match / ambiguous" model was also wrong. **Redesigned to a structural ONE-RESOURCE model.**

| Finding | Resolution |
|---|---|
| **Codex P1 + brutalist HIGH — sub-resource laundering** (`/api/accounts/{id}/settings` admitted by prefix) | **Fixed.** An endpoint is now an *item form* only if it has path_template's **exact segment count** with the collection as its non-final segments. A sub-resource (one more segment) is neither item-form nor the collection → fails closed. |
| **Codex P1 — query-routed sibling** (`/api/items?type=invoice` escaped the bound-only ?-guard) | **Fixed.** Query-routed **siblings** (non-bound) reject up front; a query-routed **bound** endpoint still defers to the existing create-collection ?-guard (blocked result `create_collection_query_routed_endpoint`, PR-D r14 preserved). |
| **brutalist HIGH (agy) — sibling collections sharing a prefix** | **Fixed** by the same segment-structural check (not string prefix): `/api/accounts-archive` ≠ `/api/accounts`. |
| **brutalist medium — deeper sub-resources under shared surface_id** | **Fixed** (segment-count exactness). |
| **CodeRabbit Major — tighten to collection-or-matched-item only** | **Fixed.** Allowed set is exactly: item-route forms (multiple forms = the same route — a pattern + sampled instances) **+** the route's collection. Nothing else. |
| **CodeRabbit Major — assert the reject signs no row** | **Fixed.** The unrelated/sub-resource/query-routed negative tests now assert `readOffensiveRunRecords(domain).length === 0`. |
| brutalist LOW — mixed relative + absolute-subdomain → multi-host reject | Intentional fail-closed (operability cost, not security). |

Also corrected a real functionality bug the ground truth exposed: a route **pattern** (`:id`) + concrete instances are the SAME route (not ambiguous) and now FIRE; `pathTemplateMatchesEndpoint` treats `:id` as non-matching, so the structural item-form check is what makes the real lab surface work.

Verified: `offensive-idor-producer` **173/173**; full `npm run test:mcp` **3182, 0 fail**.

Resolving the seven threads.

vmihalis:
@codex review
@coderabbitai review

vmihalis:
## Round-3 review addressed — `ada9bdd`

Three of the four threads are **fixed in code**; the fourth (the laundering-class residual) is **accepted + tracked as #173** per the operator decision to ship the relaxation now and close the irreducible part durably in the claims gate. Mapping:

| # | Critic(s) | Finding | Resolution |
|---|-----------|---------|------------|
| 1 | glm (design) | No min collection-depth; `/api/{id}` shallow-collapse | **Tracked #173** — same irreducible segment-semantics ambiguity as action-siblings (a shallow template treats a COLLECTION segment as the id). Bounded mitigants below; durable close = finding-subject↔`row.target` binding in the claims gate. |
| 2a | agy (security) | Action siblings (`/api/accounts/export`) pass as item-forms | **Tracked #173** — a concrete final segment is structurally indistinguishable from a slug id (`/api/accounts/alice`). Bounded, not a fake-proof. |
| 2b | Claude (security) | Shallow `path_template` collapses sibling collections | **Tracked #173** — same root cause as 1. |
| 2c | Codex (security) | `%2F` final segment classified as a valid item-form | **FIXED** — `isItemForm` now gates the final segment on the existing `capturedIdSegmentIsSafe` (the same rule `pathTemplateMatchesEndpoint` applies to the baseline) or an explicit `:id`/`{id}` route marker. `/api/accounts/foo%2Fsettings` now falls through to the `others`/`isCollection` reject instead of escaping it. New test. |
| 3 | agy + glm (correctness) | `boundEntry = items.find()||items[0]` is order-dependent (soft-block vs throw for identical content) | **FIXED** — deterministic bind: sort by pathname, then prefer a **clean (query-free) concrete** match. The bound entry is never itself query-routed, so a query sibling rejects in **either** endpoint order. New test covers both orderings (incl. the `surface.uri`-first path glm flagged as untested). |
| 4 | glm (maintainability) | Call-site comment claims a `>1-match ambiguity reject` the code never implements | **FIXED** — comment rewritten to describe the deterministic, order-independent bind (no false invariant in a MAC-signing path). |

### Why #173 is a defensible ship, not a hole

The residual is a **prose-level** mislabel risk, tightly bounded:

- `row.target` is **always** the `path_template`-derived URL — never the action endpoint or sibling collection (asserted by the existing `rows[0].target === canonicalizeExploitTarget(...)` test).
- `demonstrated_severity` is **MEDIUM-capped** by the producer.
- The derived create POST to a non-collection (e.g. `/api`) typically **fails**, so the shallow-collapse case mints **no row**.
- The **3 blind verification rounds** re-examine the finding's subject against the proof.

`claims.js:1043` already names surface↔proof-subject correctness a **PRODUCER obligation the string-binding gate cannot meet**. #173 closes it at the right layer (bind finding-subject → `row.target`), after which the producer's structural gate can be simplified.

`npm run test:mcp` 3173/0 · `check:syntax` clean · +2 IDOR tests.

vmihalis:
@codex review
@coderabbitai review

vmihalis:
## Round-4 review — all CI green; resolving the 4 re-raised threads

CI is fully green (both `test (20)` + `test (22)`, brutalist, CodeRabbit, Bob Diff Review). The `test (22)` red earlier this round was a pre-existing **flaky** OOB-poll test (#174) — it passed on re-run and on the sibling workflow on the identical commit; unrelated to this IDOR change.

The 4 remaining brutalist threads are **3 re-raises of the tracked residual + 1 maintainability nit**, merged with logged justifications per the operator's "ship now, track fix" decision:

| Critic | Finding | Disposition |
|--------|---------|-------------|
| agy (security) | Shallow `path_template` collapses `isCollection`'s guard (unrelated collections classify as item forms) | **Tracked #173** — same irreducible segment-semantics ambiguity; documented in the producer's RESIDUAL note. |
| Claude (security) | Intra-surface laundering residuals: action-sibling + shallow-template collapse classify as item forms | **Tracked #173** — explicitly the residual covered by the in-code note + issue. |
| glm (security) | Laundering prevention is contingent on `buildTargetUrl`/`row.target` in another module, not enforced by the classifier | **Tracked #173** — this is exactly the durable fix: bind the finding's subject to the proven `row.target` in the claims gate (`claims.js:1043` names it a producer obligation the string-gate can't meet). The classifier can't close it structurally; #173 closes it at the right layer. |
| glm (maintainability) | `boundEndpoint` is vestigial (one consumer) | **Justified.** Not dead: `boundEndpoint` is the dedicated input to the line-1526 bound-endpoint query-routed guard (a real, reachable soft-block when a surface records only query-bearing concrete instances — `resolveBaselineFromSurface` has already stripped the query, so the guard must inspect the *recorded* bound endpoint). Both `boundEndpoint` and `baselineUrl` are concrete item-forms of the same id-route under the same validated origin, so the safety property holds independently. The cleaner unification (derive the guard from `baselineUrl`, retire `boundEndpoint`) folds into the #173 rework, which simplifies `assertSingleHostBoundEndpoint` anyway. |

### Fixed this round (`ada9bdd`)
- Codex `%2F`: `isItemForm` now gates the final segment on `capturedIdSegmentIsSafe` / explicit `:id`/`{id}` markers.
- agy/glm correctness: deterministic, query-free-preferring bind (order-independent).
- glm maintainability: corrected the false "ambiguity reject" comment.

Merging now per the ship-now-track-fix decision; #173 carries the durable close, #174 the flaky test.

PR #175, opened 2026-06-27
fix(install): copy mcp/browser-driver.js (root cause of mass-read authed_fetch non-fire)

vmihalis:
## Root cause of the offensive mass-read producer's live non-fire

In the local lab run, `bob_http_massread_confirm` returned `blocked_by_infra` / `browser_transport_error` on **both** arms, while the general `bob_browser_*` stack worked (`bob_browser_session_start` succeeded, reached the target → 401). The lab handoff isolated it to "the producer's internal real-Chrome `authed_fetch` path."

**Why:** `scripts/install.js` copied only `["server.js", "auto-signup.js", "redaction.js"]` from top-level `mcp/`, plus `lib/*.js`. But `browser-driver.js` lives at **`mcp/browser-driver.js`** (top-level, **not** under `lib/` — it was added there in `26833e3` and never lived in `lib/`), so it was **never copied by install OR `dev-sync.sh`**.

`server.js` spawns this file at runtime as `DRIVER_SCRIPT_PATH` (`mcp/browser-driver.js`). So the operational driver was frozen on whatever first landed:
- the **older** `bob_browser_*` commands (navigate/snapshot/…) were present → general stack worked;
- the **newer** `set_auth_cookies` / `authed_fetch` commands (#155) the mass-read producer drives were **absent** → unknown-command → `browser_transport_error`.

Verified directly: `/Users/memehalis/sec/mcp/browser-driver.js` (operational) had **0** `authed_fetch`/`set_auth_cookies` occurrences; the dev driver has full handlers (cases at 479/483, impls at 669/726) and passing transport tests (`test/browser-authed-fetch.test.js` 26/0).

## Fix

- `scripts/install.js`: add `browser-driver.js` to the top-level `mcp/` copy list (+ summary line).
- `dev-sync.sh`: mirror the copy so dev workspaces refresh it too.
- `test/install-smoke.test.js`: assert `browser-driver.js` is installed — the regression guard that would have caught this (the test claimed a "complete MCP runtime" yet never checked the driver).

## Impact

This is a long-standing installer gap (since `26833e3`), not a recent regression — every install/refresh has shipped a driver that only updates when something else happens to overwrite it. After this lands, `./install.sh <target>` refreshes the driver to the current transport, and the mass-read producer's `authed_fetch` arm works.

## Tests
- `npm run test:install` 16/16, `install-smoke` 11/11, `check:syntax` clean, `bash -n dev-sync.sh` clean.

Resolves the mass-read attacker-arm `browser_transport_error` (the operational driver was stale because the installer never shipped it).

## Summary by CodeRabbit

- **Bug Fixes**
  - Fresh installs now include the browser automation runtime component as part of the complete top-level MCP runtime.
  - Reinstalls and development sync now refresh the full top-level MCP runtime set from the installer’s manifest, including `browser-driver.js`, removing stale top-level runtime files no longer included.

- **Tests**
  - Updated smoke tests to assert the installed top-level `mcp/*.js` files exactly match the installer manifest.
  - Added a reinstall regression test ensuring user-owned top-level files are preserved while expected manifest files remain.

### DIFF
--- a/dev-sync.sh
+++ b/dev-sync.sh
@@ -93,6 +93,10 @@ sync_shared_runtime() {
   cp "$SCRIPT_DIR/mcp/server.js" "$TARGET_ABS/mcp/"
   cp "$SCRIPT_DIR/mcp/auto-signup.js" "$TARGET_ABS/mcp/"
   cp "$SCRIPT_DIR/mcp/redaction.js" "$TARGET_ABS/mcp/"
+  # browser-driver.js (top-level mcp/, NOT lib/) is the Patchright subprocess server.js spawns —
+  # backs every bob_browser_* tool + the trusted authed_fetch transport. Mirror install.js so a
+  # dev-sync refreshes it too (else the driver stays frozen on whatever first landed).
+  cp "$SCRIPT_DIR/mcp/browser-driver.js" "$TARGET_ABS/mcp/"
   cp "$SCRIPT_DIR/mcp/lib/"*.js "$TARGET_ABS/mcp/lib/"
   rm -rf "$TARGET_ABS/mcp/lib/tools"
   mkdir -p "$TARGET_ABS/mcp/lib/tools"
--- a/dev-sync.sh
+++ b/dev-sync.sh
@@ -90,10 +90,12 @@ sync_shared_runtime() {
   cp "$SCRIPT_DIR/.hacker-bob/bypass-tables/"*.txt "$BOB_DIR/bypass-tables/"
 
   mkdir -p "$TARGET_ABS/mcp/lib"
-  # Glob every top-level mcp/ runtime .js (server, auto-signup, redaction, browser-driver) — a glob,
-  # NOT a hardcoded list, so a new top-level runtime file (as browser-driver.js once was — the
-  # Patchright driver server.js spawns) ships automatically instead of silently freezing the
-  # operational copy. Mirrors install.js.
+  # Copy every top-level mcp/ runtime .js (server, auto-signup, redaction, browser-driver — the
+  # Patchright driver server.js spawns; its earlier omission is what this fix addresses). dev-sync is
+  # the LOCAL test-workspace dev path, so a glob is fine here; the shipped installer
+  # (scripts/install.js) uses an explicit, test-enforced manifest instead. dev-sync intentionally does
+  # NOT fully mirror install.js — it omits e.g. lib/body-resolvers/ and the offensive-image lock,
+  # which are out of scope for a dev test workspace.
   cp "$SCRIPT_DIR/mcp/"*.js "$TARGET_ABS/mcp/"
   cp "$SCRIPT_DIR/mcp/lib/"*.js "$TARGET_ABS/mcp/lib/"
   rm -rf "$TARGET_ABS/mcp/lib/tools"
--- a/dev-sync.sh
+++ b/dev-sync.sh
@@ -90,13 +90,11 @@ sync_shared_runtime() {
   cp "$SCRIPT_DIR/.hacker-bob/bypass-tables/"*.txt "$BOB_DIR/bypass-tables/"
 
   mkdir -p "$TARGET_ABS/mcp/lib"
-  cp "$SCRIPT_DIR/mcp/server.js" "$TARGET_ABS/mcp/"
-  cp "$SCRIPT_DIR/mcp/auto-signup.js" "$TARGET_ABS/mcp/"
-  cp "$SCRIPT_DIR/mcp/redaction.js" "$TARGET_ABS/mcp/"
-  # browser-driver.js (top-level mcp/, NOT lib/) is the Patchright subprocess server.js spawns —
-  # backs every bob_browser_* tool + the trusted authed_fetch transport. Mirror install.js so a
-  # dev-sync refreshes it too (else the driver stays frozen on whatever first landed).
-  cp "$SCRIPT_DIR/mcp/browser-driver.js" "$TARGET_ABS/mcp/"
+  # Glob every top-level mcp/ runtime .js (server, auto-signup, redaction, browser-driver) — a glob,
+  # NOT a hardcoded list, so a new top-level runtime file (as browser-driver.js once was — the
+  # Patchright driver server.js spawns) ships automatically instead of silently freezing the
+  # operational copy. Mirrors install.js.
+  cp "$SCRIPT_DIR/mcp/"*.js "$TARGET_ABS/mcp/"
   cp "$SCRIPT_DIR/mcp/lib/"*.js "$TARGET_ABS/mcp/lib/"
   rm -rf "$TARGET_ABS/mcp/lib/tools"
   mkdir -p "$TARGET_ABS/mcp/lib/tools"
--- a/dev-sync.sh
+++ b/dev-sync.sh
@@ -90,15 +90,14 @@ sync_shared_runtime() {
   cp "$SCRIPT_DIR/.hacker-bob/bypass-tables/"*.txt "$BOB_DIR/bypass-tables/"
 
   mkdir -p "$TARGET_ABS/mcp/lib"
-  # Copy the top-level mcp/ runtime files from the SAME manifest the shipped installer uses
-  # (scripts/install.js MCP_TOP_LEVEL_RUNTIME_FILES) so dev and ship never skew on which files are
-  # runtime. browser-driver.js — the Patchright driver server.js spawns — is in that manifest; its
-  # earlier omission is the gap this fix closes. (dev-sync still copies only the targeted runtime set,
-  # not the full install: lib/body-resolvers/ + the offensive-image lock are out of scope here.)
-  mcp_runtime_files=$(node -e "process.stdout.write(require('$SCRIPT_DIR/scripts/install.js').MCP_TOP_LEVEL_RUNTIME_FILES.join(' '))")
-  for mcp_runtime_file in $mcp_runtime_files; do
-    cp "$SCRIPT_DIR/mcp/$mcp_runtime_file" "$TARGET_ABS/mcp/"
-  done
+  # Top-level mcp/ runtime files. scripts/install.js MCP_TOP_LEVEL_RUNTIME_FILES is the canonical,
+  # test-enforced list (install-smoke.test.js pins it == the real top-level mcp/*.js); keep these in
+  # step with it. browser-driver.js — the Patchright driver server.js spawns — is the file the installer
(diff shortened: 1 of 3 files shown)


vmihalis:
@codex review
@coderabbitai review

vmihalis:
## Round-1 review addressed — `8449474`

All three threads point at the same thing: the fix shouldn't perpetuate the hardcoded-list drift that caused the bug. Fixed at the root rather than patched per-file:

| Critic | Finding | Resolution |
|--------|---------|------------|
| glm 🟠 HIGH (maintainability) | The fix grows the hardcoded list instead of using `copyDirFiles` | **Done.** `install.js` now globs top-level `mcp/*.js` via `copyDirFiles` (which skips subdirs by design — `lib/`/`node_modules/` are copied separately). The 4 runtime files copy as before, but a future top-level file ships automatically. The list-drift that hid `browser-driver.js` is now structurally impossible. |
| glm 🟡 MEDIUM (testing) | Regression guard doesn't cover the `dev-sync` path it mirrors | **Done.** `dev-sync.sh` now uses the same `cp mcp/*.js` glob — its parallel hardcoded list is gone, so there's no separate drift to test. |
| glm 🟡 MEDIUM (testing) | Assertion guards one file's presence, not the runtime contract | **Done.** The smoke test now reads the **source** top-level `mcp/*.js` set and asserts **every** entry is installed — so dropping any runtime file (incl. the `DRIVER_SCRIPT_PATH` driver `server.js` spawns) fails the test, not just `browser-driver.js`. |

`test:install` 16/16, `bash -n dev-sync.sh` clean, `check:syntax` clean.

vmihalis:
@codex review
@coderabbitai review

vmihalis:
## Round-2 review addressed — `4185e72`

Round-1 pushed me **off** the hardcoded list **onto** a glob (HIGH); round-2 pushed back that the glob is **allow-by-default** and the test only checked `source ⊆ installed`. The two rounds pull opposite ways, so rather than pick a side I implemented the synthesis that makes both concerns moot:

| Critic | Finding | Resolution |
|--------|---------|------------|
| Claude 🟡 (×4, maintainability) | Glob is allow-by-default — a stray top-level `mcp/*.js` now ships everywhere | **Done.** Back to an **explicit manifest** `MCP_TOP_LEVEL_RUNTIME_FILES` (deny-by-default — only listed files ship). The round-1 "don't hardcode" concern is met not by a glob but by the test below. |
| agy 🟡 (testing) | Test derives its list from disk → deleting a real runtime file still passes | **Done.** Test now asserts `manifest === source top-level mcp/*.js` (deepEqual). A deletion makes source ≠ manifest → fails. |
| Claude 🟡 (testing) | Test guards only `source ⊆ installed` — can't catch the inclusion regression the glob introduced | **Done.** Test also asserts `installed top-level mcp/*.js === manifest` (deepEqual) — a dropped **or** stray file fails. |
| glm 🟡 (maintainability) | "Mirrors install.js" is false — dev-sync omits `body-resolvers/` + `offensive-image.json` | **Done.** dev-sync comment no longer claims to mirror; it states dev-sync is the loose local-dev path and deliberately doesn't fully mirror the shipped installer. |

Net: the installer copies an **explicit, deny-by-default manifest**, and the smoke test enforces `manifest === source === installed` — so the silent drift that hid `browser-driver.js` (a new runtime file forgotten in the list) now **fails CI**, and a stray file can't ship either. That's the durable resolution of the manifest-vs-glob tension, not a swing to one side.

`test:install` 16/16, `check:syntax` + `bash -n dev-sync.sh` clean.

vmihalis:
@codex review
@coderabbitai review

vmihalis:
## Round-3 review addressed — `693442f`

| Critic | Finding | Resolution |
|--------|---------|------------|
| Codex 🟡 (correctness) | Reinstall leaves a stale top-level runtime `.js` an older version shipped — "copy exactly the manifest" only held on a fresh install | **Fixed.** The installer now **removes** top-level `mcp/*.js` not in `MCP_TOP_LEVEL_RUNTIME_FILES` before copying, so a reinstall **converges** to the manifest (same `rm`-then-copy pattern install.js already uses for `lib/tools` + `lib/waves`). New test plants `mcp/legacy-driver.js`, reinstalls, asserts it's gone and the manifest files remain (17th test). This is the same stale-file class this PR exists to fix — good catch. |
| Claude 🟡 (×4, maintainability) | Dev path globs while ship path uses a frozen manifest — only one is test-enforced → dev/ship skew | **Fixed.** `dev-sync.sh` now **sources** `MCP_TOP_LEVEL_RUNTIME_FILES` from `scripts/install.js` (`node -e`) instead of globbing — one source of truth, no skew on the runtime set. |
| glm 🟡 (maintainability) | Install-summary string hardcodes the file list as prose — re-opens the drift the manifest closed | **Fixed.** The summary line now derives the list from `MCP_TOP_LEVEL_RUNTIME_FILES.join(", ")`. |

Net: **one manifest** is the single source of truth for top-level `mcp/` runtime files — consumed by the install copy + reinstall-cleanup, dev-sync, the summary, and a test pinning `manifest === source === installed` plus reinstall convergence. Nothing left to drift or skew.

`test:install` 17/17, `check:syntax` + `bash -n dev-sync.sh` clean, dev-sync manifest source resolves to the 4 runtime files.

vmihalis:
@codex review
@coderabbitai review

vmihalis:
## Round-4 review addressed — `e586418` (revert to the minimal safe fix)

Round-4 caught that my round-3 "converge to the manifest" cleanup was **destructive** — and it was right. I've reverted it. The PR is now the minimal, non-destructive fix.

| Critic | Finding | Resolution |
|--------|---------|------------|
| Codex 🟠 HIGH + glm 🟠 HIGH (correctness) | Reinstall cleanup deletes any non-manifest top-level `mcp/*.js` — destroys files Bob did not place, no backup, and the test locked it in | **Reverted.** The installer now copies the manifest files only (`copyFile` overwrites — which is what refreshes a stale `browser-driver.js`, the actual bug) and **deletes nothing**. A Bob runtime file later renamed/removed lingers harmlessly (`server.js` never `require()`s it). The destructive test is replaced by its **inverse** — "reinstall PRESERVES a user-owned top-level `mcp/` file". Ship guard is now a **superset** check (every manifest file installed), not equality, since Bob doesn't own the dir. |
| Codex 🟡 P3 + CodeRabbit 🟡 (quote the require path) | `node -e "...require('$SCRIPT_DIR/...')"` breaks / is injectable on a path with a single quote (`O'Connor`) | **Reverted.** dev-sync no longer shells a require path at all (see below). |
| glm 🟡 (maintainability) | dev-sync traded 3 `cp`s for a hard dependency on the `install.js` require-graph | **Reverted.** dev-sync is back to explicit `cp` lines (incl. `browser-driver.js`) with a comment pointing to the canonical, test-enforced manifest. |
| glm 🟡 (testing) | dev-sync (the historically-broken path) has no test | **Moot now.** The require-mechanism that needed testing is gone — dev-sync is 4 literal `cp` lines. The ship side stays test-enforced (`manifest === source`, every manifest file installed). |

Net: copy Bob's runtime files (overwriting stale ones), **never touch the user's**. The genuinely useful anti-drift mechanism — the manifest + the `manifest === source` test that would have caught the original `browser-driver.js` omission — is kept; the over-engineered cleanup/require-sourcing is gone.

`test:install` 17/17 (incl. the user-file-preservation test), `check:syntax` + `bash -n` clean.

vmihalis:
@codex review
@coderabbitai review

vmihalis:
## Round-5 review addressed — `a42a445` (test-only)

Both findings were legitimate coverage gaps, now closed:

- **CodeRabbit (Major) + glm — assert refresh, not just presence:** the reinstall test now seeds a STALE `browser-driver.js`, reinstalls, and asserts the stale content is gone AND the file equals the current source. That tests the *actual* frozen-driver bug (`copyFile` overwrites on reinstall), not just file existence.
- **agy (×4) — dev-sync/manifest drift:** new test pins them — every `MCP_TOP_LEVEL_RUNTIME_FILES` entry must appear as a `dev-sync.sh` cp target, so adding a runtime file without updating dev-sync fails CI. Drift caught by a test instead of coupling dev-sync to the install.js require-graph (the fragile, quote-injectable approach reverted in round 4).

No runtime change this round. The fix now has: ships the driver, REFRESHES a stale one on reinstall, PRESERVES user files, `manifest === source`, and dev-sync parity. `test:install` 18/18.

vmihalis:
@codex review
@coderabbitai review

vmihalis:
## Merging with logged justification (round 6)

All 7 CI checks are green (6th consecutive round). The 3 remaining threads are **advisory brutalist MEDIUMs**, all about the dev-sync DRY question, which has now cycled through every option with contradictory feedback each round:

- round-2: glob → "allow-by-default, ships strays"
- round-3: `node -e require(install.js)` → "fragile require-graph + single-quote path injection"
- round-4: revert to explicit `cp` → "dev/ship skew"
- round-5: add a parity test → round-6: "substring search too weak / manifest not consumed at runtime"

There is no approach to a **bash dev-only tool** consuming a **JS constant** that satisfies all of these at once — they're mutually exclusive. Judging each remaining finding:

1. **Codex — "manifest not consumed by dev-sync (two parallel lists)":** Consuming it at runtime is what round-3 did and round-4 correctly rejected (fragile + injectable). A bash tool can't cleanly import a JS const. The drift is guarded by a test, which is the right layer. **Accepted tradeoff for a dev-only convenience script.**
2. **glm — "renamed/removed file harmlessness untested":** The harmlessness is true by construction — `server.js` `require()`s a fixed set, so an unreferenced lingering file is never loaded. The alternative (delete by negation) was correctly reverted as destructive in round 4; metadata-based pruning is over-engineering for a cosmetic, harmless case. **Won't-fix, by design.**
3. **glm ×4 — "parity test is a substring search":** Conceded as a minor test-quality preference. The current check (`mcp/<name>` must appear in `dev-sync.sh`) does catch the real drift case (a manifest runtime file with no dev-sync line); the `mcp/` prefix ties it to the cp target, not the prose comment. A regex would be marginally tighter but doesn't change what's caught. **Adequate.**

The shipped fix — the part that matters — is correct and thoroughly tested: it copies `browser-driver.js` (the frozen-driver root cause), **refreshes** a stale copy on reinstall (asserted), **preserves** user-owned files (asserted), and the `manifest === source` drift guard would have caught the original omission. None of the 3 MEDIUMs affects that.

Per the repo's merge rule (advisory findings may be merged with a clear logged justification when judged low-value/contradictory), merging now.

